PR #102, opened 2014-11-02
Django REST Framework support

### LINKED ISSUE #99: Django-Rest-Framework Integration instead of Tastypie
Django rest framework is the most populer rest framework for django. how to use django notif with that?

brianlittmann:
Closes #99

### DIFF
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -8,6 +8,8 @@ from rest_framework.viewsets import ModelViewSet
 
 from push_notifications.models import APNSDevice, GCMDevice
 
+HEX64_RE = re.compile("[0-9a-f]{64}", re.IGNORE_CASE)
+
 
 # Serializers
 class APNSDeviceSerializer(ModelSerializer):
@@ -16,9 +18,8 @@ class APNSDeviceSerializer(ModelSerializer):
 
 	def validate_registration_id(self, attrs, source):
 		# iOS device tokens are 256-bit hexadecimal (64 characters)
-		hex64re = re.compile("[0-9a-fA-F]{64}")
 
-		if hex64re.match(attrs[source]) is None:
+		if HEX64_RE.match(attrs[source]) is None:
 			raise ValidationError("Registration ID (device token) is invalid")
 		return attrs
 
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -16,7 +16,9 @@ class APNSDeviceSerializer(ModelSerializer):
 
 	def validate_registration_id(self, attrs, source):
 		# iOS device tokens are 256-bit hexadecimal (64 characters)
-		if re.match("[0-9a-fA-F]{64}", attrs[source]) is None:
+		hex64re = re.compile("[0-9a-fA-F]{64}")
+
+		if hex64re.match(attrs[source]) is None:
 			raise ValidationError("Registration ID (device token) is invalid")
 		return attrs
 
@@ -39,11 +41,12 @@ class IsOwnerOrReadOnly(permissions.BasePermission):
 # Mixins
 class DeviceMixin(object):
 	def pre_save(self, obj):
-		if self.request.user.is_authenticated(): obj.user = self.request.user
+		if self.request.user.is_authenticated():
+			obj.user = self.request.user
 
 
 class AuthorizedMixin(object):
-	permission_classes = (permissions.IsAuthenticated, IsOwnerOrReadOnly,)
+	permission_classes = (permissions.IsAuthenticated, IsOwnerOrReadOnly)
 
 
 # ViewSets
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -8,7 +8,7 @@ from rest_framework.viewsets import ModelViewSet
 
 from push_notifications.models import APNSDevice, GCMDevice
 
-HEX64_RE = re.compile("[0-9a-f]{64}", re.IGNORE_CASE)
+HEX64_RE = re.compile("[0-9a-f]{64}", re.IGNORECASE)
 
 
 # Serializers
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -1,0 +1,65 @@
+from __future__ import absolute_import
+
+import re
+
+from rest_framework import permissions
+from rest_framework.serializers import ModelSerializer, ValidationError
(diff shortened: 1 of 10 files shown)


push_notifications/api/rest_framework.py
```diff
@@ -0,0 +1,69 @@
...
+			obj.user = self.request.user
+
+
+class AuthorizedMixin(object):
+	permission_classes = (permissions.IsAuthenticated, IsOwnerOrReadOnly)
+
+
+# ViewSets
+class APNSDeviceViewSet(DeviceMixin, ModelViewSet):
+	queryset = APNSDevice.objects.all()
```
sayplastic: from [DRF Generic Views documentation](http://www.django-rest-framework.org/api-guide/generic-views/):
> self.queryset gets evaluated only once, and those results are cached for all subsequent requests.
So it would be better to redefine get_queryset method here

brianlittmann: I think that is only applicable for overriding methods in the [GenericAPIView](http://www.django-rest-framework.org/api-guide/generic-views/#genericapiview). Regarding the `queryset` attribute:

> If you are overriding a view method, it is important that you call `get_queryset()` instead of accessing this property directly

But here we're using a [ModelViewSet](http://www.django-rest-framework.org/api-guide/viewsets/#modelviewset), where the docs state:
> ...you'll normally need to provide at least the `queryset` and `serializer_class` attributes

and the `GenericAPIView` method `get_queryset` simply returns that queryset:

> Defaults to returning the queryset specified by the `queryset` attribute

I may be missing something though. Could you provide some code of what you recommend? Here is the DRF example of a `ModelViewSet`:

```python
class AccountViewSet(viewsets.ModelViewSet):
    """
    A simple ViewSet for viewing and editing accounts.
    """
    queryset = Account.objects.all()
    serializer_class = AccountSerializer
```

danielbraun:
The suggested PR doesn't work with a current DRF version (3.1.0). It needs a rewrite.
For example, `pre_save` is not compatible with DRF > 3.0.

brianlittmann:
Thanks, @danielbraun. It looks like @jleclanche prefers the other PR #105, otherwise I will update this one to be compatible with DRF 3.*.

danielbraun:
The other one isn't 100% compatible as well. I'll make a PR, or even a separate repository (I'm not sure it should be a part of this repo anyway) once I'm done integrating it into my app.



‫ב-22 באפר׳ 2015, בשעה 17:20, ‏‏Brian Littmann ‏<notifications@github.com> כתב/ה:‬

jleclanche:
I'm fine with either approach to be honest, and I'm cool with having it in the repository as long as it's an optional dep.

matthewh:
What is the status of this or the #105 PR? I'd like to help get this finished!

chrisballinger:
I'd also like to help see this merged in some way.

jleclanche:
I need something that's been reviewed and tested by a couple of people at least and that passes travis (and adds tests if possible). It can't add a hard dep either - should work like the tastypie one.

matthewh:
Great, I integrated this into a private branch and and made some trivial adjustments for DRF 3.x support. Otherwise it works well so far. I'll comment again once I've completed my testing.

jleclanche:
@matthewh Can you file a new PR with what you have, with the commits appropriately cleaned up and squashed if necessary? I'm having a lot of trouble reviewing what's here.

awidgery:
@matthewh Would love to use this - how's your testing going?

matthewh:
I have it working with v3 - need to make sure it still works with v2. One detail I'm not comfortable with in this version is that the tokens are exposed as readable to all users. I'd love to hear from others if they think this is a useful feature or security problem.

gmorell:
I'd feel that if you really wanted the tokens not to be exposed, the end user can just subclass the serializers to show/hide the fields they want.

jleclanche:
I'm ok with a version that only supports v3, tbh.

chrisballinger:
@matthewh @gmorell Why not secure by default?

matthewh:
+3 for better security - I'll restrict the default view to just the tokens uploaded by the current user.

awidgery:
+1 for a version that only supports v3 - I think that's fine.

azevakin:
For DRF v3
```python
from rest_framework import serializers
from push_notifications.models import GCMDevice, APNSDevice


class CurrentUserMixin(object):
    user = serializers.PrimaryKeyRelatedField(
        default=serializers.CurrentUserDefault(),
        read_only=True)


class GCMDeviceSerializer(CurrentUserMixin, serializers.ModelSerializer):

    class Meta:
        model = GCMDevice


class APNSDeviceSerializer(CurrentUserMixin, serializers.ModelSerializer):

    class Meta:
        model = APNSDevice


class DeviceMixin(object):

    def get_queryset(self):
        queryset = super(DeviceMixin, self).get_queryset()
        return queryset.filter(user=self.request.user)
```

danielbraun:
Just posting what I ended up with, for reference:
```
from rest_framework.viewsets import ModelViewSet
from rest_framework.serializers import ModelSerializer
from push_notifications.models import APNSDevice, GCMDevice


class DeviceSerializerMixin(object):

    class Meta:
        fields = ('name', 'device_id', 'registration_id', 'date_created')


class APNSDeviceSerializer(DeviceSerializerMixin, ModelSerializer):

    class Meta(DeviceSerializerMixin.Meta):
        model = APNSDevice


class GCMDeviceSerializer(DeviceSerializerMixin, ModelSerializer):

    class Meta(DeviceSerializerMixin.Meta):
        model = GCMDevice


class DeviceViewSetMixin(object):
    lookup_field = 'registration_id'

    def get_queryset(self):
        return self.queryset.filter(user=self.request.user)

    def perform_create(self, serializer):
        serializer.save(user=self.request.user)


class APNSDeviceViewSet(DeviceViewSetMixin, ModelViewSet):
    queryset = APNSDevice.objects.all()
    serializer_class = APNSDeviceSerializer


class GCMDeviceViewSet(DeviceViewSetMixin, ModelViewSet):
    queryset = GCMDevice.objects.all()
    serializer_class = GCMDeviceSerializer
```

azevakin:
@danielbraun perform_create only for DRF v2

matthewh:
@azevakin Please read http://www.django-rest-framework.org/topics/3.0-announcement/#generic-views. @danielbraun is correct, perform_create is for DRF v3.

azevakin:
@matthewh for some reason I thought that it was about pre_save

jleclanche:
Closing in favour of #160

PR #105, opened 2014-11-06
Django REST Framework (alt)

peterhinson:
Optional serializers, view sets, and URL routes for use with Django REST Framework. The included views allow authenticated users to view/add/remove devices associated with their account. These views can be used as is or inherited from if you'd like to modify permissions or functionality.

Pretty simple to use. Here's an example of how to include the URL routes:

```python
from push_notifications.django_rest_api import gcm_list, apns_list, gcm_detail, apns_detail

# ... in your URL patterns

# GCM API notifications
url(r'^api/devices/gcm/$',
    gcm_list,
    name='api_devices_gcm_list'),

url(r'^api/devices/gcm/(?P<registration_id>[\w-]+)/$',
    gcm_detail,
    name='api_devices_gcm_detail'),

# APNS notifications
url(r'^api/devices/apns/$',
    apns_list,
    name='api_devices_apns_list'),

url(r'^api/devices/apns/(?P<registration_id>[\w-]+)/$',
    apns_detail,
    name='api_devices_apns_detail'),
```

### DIFF
--- a/push_notifications/django_rest_api.py
+++ b/push_notifications/django_rest_api.py
@@ -1,0 +1,74 @@
+from rest_framework import serializers, permissions, viewsets
+from .models import APNSDevice, GCMDevice
+
+
+class IsOwner(permissions.BasePermission):
+    """
+    Custom permission to only allow owners of an object to see it.
+    """
+
+    def has_object_permission(self, request, view, obj):
+        # Only allow the user to see their own devices
+        return obj.user == request.user
+
+
+class APNSDeviceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = APNSDevice
+        fields = ('name', 'device_id', 'registration_id')
+
+    read_only = ('date_created',)
+
+
+class GCMDeviceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = GCMDevice
+        fields = ('name', 'device_id', 'registration_id')
+
+    read_only = ('date_created',)
+
+
+class DeviceViewSetMixin(object):
+    lookup_field = 'registration_id'
+    queryset = APNSDevice.objects.all()
+    serializer_class = APNSDeviceSerializer
+    permission_classes = (permissions.IsAuthenticated,
+                          IsOwner,)
+
+    def pre_save(self, obj):
+        # Set the requesting user as the device user
+        obj.user = self.request.user
+
+
+class APNSDeviceViewSet(DeviceViewSetMixin, viewsets.ModelViewSet):
+    queryset = APNSDevice.objects.all()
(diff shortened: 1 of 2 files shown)


bender314:
Hey guys. It's nice to see that someone did the job before I started. I just wanted to add some notes:
1. pre_save won't work with DRF 3.0. You need to use perform_create and perform_update instead. I think it would be best to provide both of those methods to be compatible with both versions. Reference: http://www.django-rest-framework.org/topics/3.0-announcement/
2. This entire app breaks pep8 by using wrong indentations (tabs). This pull request uses 4 spaces which is a valid indentation but inconsistant with the rest of the project. I would suggest using tabs for this pull request and tell the project author to run autopep8 on the entire project.
3. I don't understand why the other solution offers read only access without filtering by request.user. What is the usage scenario for that? For me it seems reasonable to filter this out for all requests but I'm implementing push notifications for the first time and I don't really know how it works yet.

@brianlittmann As weird as it is the device_id doesn't need to be unique http://stackoverflow.com/a/2338839

bender314:
Also the registration_id is not unique for the GCMDevice which makes it bad lookup_field (if conflict is created it will raise MultipleObjectsReturned for the detail view)

jleclanche:
Guys, I don't use Django-Rest-Framework so I would love some feedback on which approach is preferred by users between #102 and #105.

I'd love to get this merged and release 1.3.

jleclanche:
For what it's worth, code-wise, I prefer #105 and possibly expanding on it later.

push_notifications/django_rest_api.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class DeviceViewSetMixin(object):
+    lookup_field = 'registration_id'
+    queryset = APNSDevice.objects.all()
+    serializer_class = APNSDeviceSerializer
+    permission_classes = (permissions.IsAuthenticated,
+                          IsOwner,)
+
+    def pre_save(self, obj):
```
danielbraun: Doesn't work with DRF > 3.0

push_notifications/django_rest_api.py
```diff
@@ -0,0 +1,74 @@
...
+    class Meta:
+        model = GCMDevice
+        fields = ('name', 'device_id', 'registration_id')
+
+    read_only = ('date_created',)
+
+
+class DeviceViewSetMixin(object):
+    lookup_field = 'registration_id'
+    queryset = APNSDevice.objects.all()
```
danielbraun: Is it right? GCM Also inherits from this, why does it only refer to APNSDevices?

jleclanche:
Closing in favour of #102.

PR #135, opened 2015-03-24
Fix Django 1.3 compatibility issue

enikesha:
`./manage.py migrate` was failing with
`db_type() got multiple values for keyword argument 'connection'`
due to a way `call_with_connection` is implemented in `django/db/models/fields/subclassing.py` in Django 1.3


jleclanche:
Thank you!

Note that Django 1.3 won't be supported after the next release.

PR #137, opened 2015-03-29
django 1.8 compatibility fixes

mochawich:
- removed unnecessary import causing import error 
- added another migration file to use django uuid or ones from uuidfield package


jleclanche:
This should no longer be necessary with f21a4d46769389c27a2ff8e31ba10340943aa2a1

PR #138, opened 2015-04-01
Error causing broken socket during bulk apns

gesinger:
Reported in https://github.com/jleclanche/django-push-notifications/issues/122 , when sending bulk APNS messages, Apple will make the socket available for reading after an issue with one of the IDs. Writing to the socket may not do anything afterwards, and Apple requires resending that notification as well as all notifications sent after it. In https://developer.apple.com/library/ios/technotes/tn2265/_index.html#//apple_ref/doc/uid/DTS40010376-CH1-TNTAG44

```
"Here's how to check for errors when using the enhanced notification format. Keep writing until a write fails. If the stream is ready for writing again, resend the notification and keep going. If the stream isn't ready for writing, see if the stream is available for reading.

If it is, read everything available from the stream. If you get zero bytes back, the connection was closed because of an error such as an invalid command byte or other parsing error. If you get six bytes back, that's an error response that you can check for the response code and the ID of the notification that caused the error. You'll need to send every notification following that one again."
```

The approach used here (to make it easier) is to simply wait a length of time to see if the socket is available for reading, and reset the socket if it is. We ignore the message, and skip the PN that caused the error. In testing, I've found that 100ms (an APNS_BULK_ERROR_CHECK_DELAY_SECONDS of 0.1) seems to be enough time, however, I set it to 0 so that current users will not see any delay in their message sending.


gesinger:
Another alternative would be to check to see if the socket is available for reading (with a small timeout), and continue if the timeout hits and it isn't available, with a larger timeout at the end, and if the socket is available for reading, read the 6 bytes and resend all notifications starting at the identifier sent by Apple. This would afford a smaller timeout on each send (though we'd have another setting for the final timeout).

jleclanche:
Code looks good (can you squash please?), but I'm not sure this is the correct solution for apns. Would need someone more familiar with it to review.

gesinger:
Thanks for taking a look. I'll probably take another look at it too. In some further testing, due to changing response times, the timeout would have to be somewhat high for someone sending a lot of messages. I think something closer to the second approach (checking the response and resending) is probably a better real world solution. It would result in faster individual sends, just a higher cost of resending if some have failures (versus this approach which has higher delay costs, but no resending). I'd definitely like to hear other approaches as well.

gesinger:
I have a new proposal that I believe will be a better solution. Reading through Apple's technical documents, they point out that it is faster to simply send all the messages, then check for the error after (as the cost of resending all of the messages is fairly cheap).

Since there's already logic that checks for the errors at the end of the send block and raises an APNSServerError with the identifier in the event of a failure, we should catch it in the bulk send block and resend that identifier + 1 (with a new socket). All of the work is pretty much there, it just shifts the responsibility of the resends onto django-push-notification instead of the user (the user would have to find the offending device and mark it as inactive before continuing otherwise).

gesinger:
Created https://github.com/jleclanche/django-push-notifications/pull/139

PR #144, opened 2015-04-25
makemigrations resulting in additional migration

mochawich:
db_index=True is missing from both devices making the 'makemigrations' command result in extra migration inside the package itself.


jleclanche:
Please rebase your commits

mochawich:
done.

PR #146, opened 2015-04-30
Add retry logic to bulk APNS messages when Apple reports a failing ident...

pcompassion:
https://github.com/jleclanche/django-push-notifications/pull/139

"@gesinger Can you fix travis and rebase on top of master, please?"
I did what you asked for @gesinger

Not sure if it's ok to do so..


jleclanche:
He already rebased, but I need people to test it and report before I land it. I use the lib very little nowadays so I rely on the users.

PR #154, opened 2015-06-05
Update vs for loop on prune devices

gio82:
Updating should be faster than for loop.

Not tested.


jleclanche:
Please squash the commits.

jleclanche:
@gio82 you don't need to close the PR to squash. you can just `git push --force` to it.

PR #156, opened 2015-06-06
APNS bulk retry on fail

gio82:
Still not the best solution but gets the job done.

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -51,10 +51,7 @@ class DeviceAdmin(admin.ModelAdmin):
 		# could very easily leave an expired device as active.  Maybe
 		#  this is just a bad API.
 		expired = get_expired_tokens()
-		devices = queryset.filter(registration_id__in=expired)
-		for d in devices:
-			d.active = False
-			d.save()
+		queryset.filter(registration_id__in=expired).update(active=False)
 
 
 admin.site.register(APNSDevice, DeviceAdmin)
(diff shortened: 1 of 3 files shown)


jleclanche:
I'm not a fan of it but I'll merge it if it gets tested and approved.

gio82:
Honestly I don't like it either. 
Maybe the best solution, avoiding to reinvent the wheel, would be to use PyAPNs libraries.

jleclanche:
That's a huge dependency as it depends on Twisted. I'm not willing to let that go in just for this.

jleclanche:
@gio82 please squash your commits

gio82:
I made major amendments on the way APNS and GCM return results when bulk send message.

In particular both of them returns as common pattern:

    {'failure': 0, 'success': 0, 'invalid_registration_ids': []}

This way it's known how many failures, successes and where the failures occur.
`invalid_registration_ids` may be use to set `active=False` or directly `delete()`

GCM returns an additional key that is `canonical_ids`.

gio82:
Actually I'm still working on GCM error messages. I'll update this code soon

singleton11:
I think resending depends from buisness logic of project, because there is can be more than 1000 devices, and we need to answer to following questions:
1. through how many devices we should check errors (we need at least one second for error checking). If we have one error checking per 100 devices, we need 10 seconds to sending these messages.
2. What timeout should be set.
There is should be developer pain, but not django-push-notification. But I can mistake

jleclanche:
Closing this, no longer relevant with pyapns2

PR #158, opened 2015-06-09
Parse gcm server response and set device active = false if either NotRegistered or InvalidRegistration errors are returned.  All other errors will continue to throw NotificationError

jamaalscarlett:


jleclanche:
Your logic is destructive and doesn't check for what kind of error happens. That means if you mistakenly send, for example, an invalid TTL you will be deleting all the IDs it was sent to.

jamaalscarlett:
I am checking for the error message.  Only NotRegistered or InvalidRegistration are affected, and the device is de-activated, not deleted.

jleclanche:
No? You're only checking in the tests...

jamaalscarlett:
I don't understand the test is being run against gcm.py, that is where the checking logic resides.

jamaalscarlett:
I will add tests for other error codes to verify.

jamaalscarlett:
Added a test to raise GCMError

PR #161, opened 2015-06-12
Toxify the project

ticosax:
Test against all supported python versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ build
 
 # virtualenv
 .env
+
+# tox
+.tox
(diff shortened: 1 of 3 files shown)


jleclanche:
Looks good, but is there any reason I need this?

ticosax:
At the moment only to test `locally` will all python versions.
If later you want to support several django versions, tox will become your best friend

ticosax:
regarding `requirements.txt` and tox. It is not always a good idea to couple them.

Because in tox you iterate over different exact version of your dependencies, while in `requirements.txt` you normally soft pin the version like `Django>=1.8`

jleclanche:
Fair enough.

jleclanche:
Landed in cfd9ae0.

Issue #68, opened 2014-08-05
GCM device id displayed as integer in admin


borjafp:
When edit a GCM device record the id value is displayed as integer but the field require hex value to validate the form. 

You have to convert the integer value in hex manually before save the record.

jleclanche:
Is this still an issue for everyone on the latest versions of django and uuidfield?

borjafp:
Yes, tested on django 1.7.4 and uuidfield 0.5.0.

jleclanche:
Would be great to test on 1.8 with the changes I just pushed.

jleclanche:
Version 1.2.1 is out, please try it in there.

jamaalscarlett:
Still exists in latest build.
See https://github.com/jleclanche/django-push-notifications/pull/163

jleclanche:
This is fixed in 1.3

PR #163, opened 2015-06-16
Fixed bug #68. Hex or Integer values can be used when modifying existi…

jamaalscarlett:
…ng GCMDevice device_id field.

### DIFF
--- a/push_notifications/fields.py
+++ b/push_notifications/fields.py
@@ -10,7 +10,7 @@ from django.utils.translation import ugettext_lazy as _
 __all__ = ["HexadecimalField", "HexIntegerField"]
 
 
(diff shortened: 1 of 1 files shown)


jleclanche:
This isn't the right fix. The field shouldn't be displayed as an int in the admin in the first place.

jamaalscarlett:
Hmm.  This may not be worth the trouble.  Changing it in the admin would require implementing from_db_value().  While that works, it means the value returned from the db is always represented as a hex value, not the integer value as it was previously.  I am using django rest framework so my serializer is now passing back "0x1" instead of "0" in my api call.  If I had any android code that was calling JSONObject.getInt() on the value assuming an int, it would now throw an exception.

PR #168, opened 2015-06-23
Add support of python3 for prune_devices management command

ticosax:
Under python 3 `rune_devices management command` is failing with the folowing error

```
'bytes' object has no attribute 'encode'
 push_notifications/apns.py, apns_fetch_inactive_ids:234
```
Let's address this issue by using `codecs` module.


jleclanche:
Looks good. Merging if travis passes.

PR #176, opened 2015-06-30
Add missing migration for python 3

### LINKED ISSUE #175: Makemigrations creates a new migration when running python 3
Running `python manage.py makemigrations` on a python 3 project creates the following migration:

```python
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import models, migrations
import push_notifications.fields


class Migration(migrations.Migration):

    dependencies = [
        ('push_notifications', '0001_initial'),
    ]

    operations = [
        migrations.AlterField(
            model_name='apnsdevice',
            name='device_id',
            field=models.UUIDField(blank=True, null=True, verbose_name='Device ID', help_text='UDID / UIDevice.identifierForVendor()', db_index=True),
        ),
        migrations.AlterField(
            model_name='gcmdevice',
            name='device_id',
            field=push_notifications.fields.HexIntegerField(blank=True, verbose_name='Device ID', help_text='ANDROID_ID / TelephonyManager.getDeviceId() (always as hex)', null=True, db_index=True),
        ),
    ]
```

This does not happen on python 2.

I suspect this is caused by the presence of `bytes` in the [initial migration](https://github.com/jleclanche/django-push-notifications/blob/master/push_notifications/migrations/0001_initial.py#L23) (and [here](https://github.com/jleclanche/django-push-notifications/blob/master/push_notifications/migrations/0001_initial.py#L39)).

Ian-Foote:
Fixes #175.

### DIFF
--- a/push_notifications/migrations/0002_python3_help_text.py
+++ b/push_notifications/migrations/0002_python3_help_text.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import push_notifications.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('push_notifications', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='apnsdevice',
+            name='device_id',
+            field=models.UUIDField(blank=True, null=True, verbose_name='Device ID', help_text='UDID / UIDevice.identifierForVendor()', db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='gcmdevice',
+            name='device_id',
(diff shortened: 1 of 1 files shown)


jleclanche:
What does that migration do? it looks dummy.

Ian-Foote:
In the initial migration, the [`help_text`](https://github.com/Ian-Foote/django-push-notifications/blob/fix-python3-migrations/push_notifications/migrations/0001_initial.py#L23) for `device_id` is explicitly marked as `bytes`. This updates that to the native string type - still bytes on python 2 but unicode on python 3.

jleclanche:
Just need to change the initial migration then; the help text is not in the db...

jamaalscarlett:
I have verified that a change to the initial migration fixes the issue for both python 2 and 3

jleclanche:
@jamaalscarlett This is no longer needed right? Or was the change not made?

jamaalscarlett:
I haven't fixed the original migration yet, but this new migration is not needed.

PR #178, opened 2015-07-03
Enabled python_2_unicode_compatible __str__ method for Device model. …

### LINKED ISSUE #117: Port 'Device' Model `__unicode__()` Method to Python 3 (?)
The `Device` model implements a `__unicode__()` method that breaks with python 3, for example, in the admin app:
image

Possible, backwards-compatible solution to this is implementing `__str__()` while using Django's `@python_2_unicode_compatible`:
https://docs.djangoproject.com/en/dev/topics/python3/#str-and-unicode-methods

matthewh:
Models will now be displayed correctly with print in py2 and py3. Fixes #117.

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -8,7 +8,7 @@ User = get_user_model()
 
 
 class DeviceAdmin(admin.ModelAdmin):
-	list_display = ("__unicode__", "device_id", "user", "active", "date_created")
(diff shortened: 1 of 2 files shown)


jamaalscarlett:
You need to change the references to __str__ in admin.py

matthewh:
@jamaalscarlett I updated the __unicode__ references in admin.py to __str__.

PR #180, opened 2015-07-04
Add support for http proxy when connecting to APNS

BastiPfa:
Hi,

if your server is located behind a firewall and you're not able / not allowed or simply don't want to configure your firewall to allow a direct socket connection to the Apple Push Notification Service you may have to / want to use a proxy. Otherwise the connection attempts will fail.

Added support for this use case using the PySocks package.

Best regards,
Basti

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 env:
   - TOXENV=flake8
-  - TOXENV=py27-django18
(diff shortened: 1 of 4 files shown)


jleclanche:
Why are you closing your own PRs?

You can update them by force-pushing to your own branch.

PR #181, opened 2015-07-04
Add support for http proxy when connecting to APNS

BastiPfa:
Hi,

if your server is located behind a firewall and you're not able / not allowed or simply don't want to configure your firewall to allow a direct socket connection to the Apple Push Notification Service you may have to / want to use a proxy. Otherwise the connection attempts will fail.

Added support for this use case using the PySocks package.

Best regards,
Basti

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 env:
   - TOXENV=flake8
-  - TOXENV=py27-django18
-  - TOXENV=py33-django18
-  - TOXENV=py34-django18
+  - TOXENV=py27-django18-pysocks154
+  - TOXENV=py33-django18-pysocks154
+  - TOXENV=py34-django18-pysocks154
 install:
   - pip install tox
 script:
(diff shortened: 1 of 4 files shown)


matthewh:
May I suggest making PySocks package optional and use a [django system check](https://docs.djangoproject.com/en/1.8/ref/checks/) to warn the user to install the package if a proxy is present?

BastiPfa:
I will have a look at that possibility in the upcoming days.

BastiPfa:
Nobody seems to be interested in this feature so I'll close the Pull Request

jleclanche:
Sorry about this, I didn't notice the PR and I haven't been able to work on the project myself.

Adding a dependency for a feature few will use is not an option. @matthewh's suggestion would have to be put into place for this PR to be accepted.

However I'm surprised any change in the code is necessary for a proxy to work. Why APNS but not GCM?

BastiPfa:
Absolutely no problem.
I just wanted to provide this workaround if anyone has the same problem as I had that the firewall is blocking direct socket connections from the internal network to the APNS service.

With GCM everything worked fine in my case, APNS socket connections were blocked...

jleclanche:
I'm also wondering if it's wise to use the `$https_proxy` variable. APNS isn't over https.

Issue #130, opened 2015-03-18
Can we use different certificates/api keys?


egamonal:
in order to send a message to APN and GCM we need to add the settings `APNS_CERTIFICATE`, etc. 
it looks like I can't change the certificate before sending the notifications. is that correct?

my use case is a dashboard where each user has a mobile app that will receive the notifications. these apps are completely independent and can't use the same certificate.

jleclanche:
The ability to use multiple certificates (like django's DATABASES/CACHES dicts) has been brought up before, I can't find it. I'm not opposed to it... I dislike breaking backwards compatibility, but if this proves to be a common request, I'm willing to let it in.

jackfengji:
I also want this feature in my project. I was thinking just add a "key" parameter to the functions. And use that or the default one if it's not provided.

evdoks:
We will be needing this feature as well. We will start working on these extension for our project, so it might be a good place to discuss the design, in case we agree on letting this extension in. Actually, it does not have to break backwards compatibility - we are thinking about storing multiple certificates in a database table, fallbacking to `settings` if the table is empty.

jleclanche:
The certs should most definitely not be in the db.

There should be a setting similar to the current one but presenting a dict that works like django's DATABASES setting.
The current setting should be backwards compatible but raise a deprecation warning.

evdoks:
I don't think it will scale well if certificates are to be managed via a settings file. We are planning to use 10+ different applications, so managing all the certificates will quickly become quite a challenge. You do not want to store certificates in the repo, so it is already tricky enough to manage just one of them. What is your concert with keeping them in the database? [django-ios-notifications](https://github.com/stephenmuss/django-ios-notifications) is doing it this way and it seems to be working pretty well. As I said, backward compatibility should not be an issue.

jleclanche:
Putting settings in the database prevents people from being able to manage their settings by file, simple as that.

If you want it to scale, you can write a settings loader of sorts, seeing as settings are just python file. i don't see the issue. 10 certs, 100 certs, all the same.

t-io:
I see your point for manage settings by file. But unfortunately that's not always possible. For example when you use some PaaS hosting like heroku. I see that it would work with only a few certificates set and extended via environment variables but this scales poor.

jleclanche:
I have a very, very solid -2 on depending on db settings here.

niktto:
[...]

PR #182, opened 2015-07-05
Multiple apns certificates

cristiano2lopes:
Initial work on #130.

Documentation and test for the added features were added.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,7 +46,12 @@ Edit your settings.py file:
 
 	PUSH_NOTIFICATIONS_SETTINGS = {
 		"GCM_API_KEY": "<your api key>",
-		"APNS_CERTIFICATE": "/path/to/your/certificate.pem",
+		"APNS_APP_CERTIFICATES": {
+			"YOUR_APP_NAME": {
+				"APNS_CERTIFICATE": "/path/to/your/certificate.pem",
+			}
+		}
+		
 	}
 
 .. note::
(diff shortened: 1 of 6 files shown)


jleclanche:
I'll have to do a more in-depth review once I get home, however a quick heads up. I'm seeing a lot of code that checks for deprecated settings and such - I'd like to move instead to writing Django 1.7 system checks. See this page for reference:

https://docs.djangoproject.com/en/1.8/ref/checks/

Here's an example of how we use them in Pootle:

https://github.com/translate/pootle/blob/master/pootle/checks.py

cristiano2lopes:
Will change the test files to /dev/null. 
Regarding the checks I'am not familiar yet with the most recent releases of django (in fact I had to bac port this to the 1.2 release for work) but I can look into it. There is for sure a better way to do that on 1.7 and 1.8.

matthewh:
What else is needed before this can be merged?

jleclanche:
@matthewh I'd like to see the checks changed, and then i can do another more thorough review

matthewh:
@jleclanche I rewrote the inline deprecated setting checks using django system checks but I see one issue with this change that makes me lean towards leaving it as is. It presumes the developer will always have APNS configured but it should still be possible to just use GCM. Unless we just make all of the checks warnings... What do you think?

jleclanche:
You can simply skip the checks if there's nothing in the `APNS` settings dict.

jleclanche:
Superceded by #226

PR #189, opened 2015-07-23
Fixed bug that caused an error when creating or updating a gcm device

jamaalscarlett:
When either postgres or mysql is selected as the database engine
attempts to create or update a gcm device using a form and http post
will fail.  The validators were being run against a string value
thus min and max validation were failing.


jamaalscarlett:
This should fix issues https://github.com/jleclanche/django-push-notifications/issues/187 and https://github.com/jleclanche/django-push-notifications/issues/171

jamaalscarlett:
This should fix issues https://github.com/jleclanche/django-push-notifications/issues/171 and https://github.com/jleclanche/django-push-notifications/issues/187

Issue #192, opened 2015-07-24
Error in GCMDeviceAdmin


alfaro28:
I was trying to edit some GCMDevices in the admin but I was getting the exception

    hex() argument can't be converted to hex
    at push_notifications/admin.py in device_id_hex, line 66

I think this was an issue because my device_id was NULL, I think a simple null check would solve the issue.

My temporary fix was to put this in my own admin.py file:

    admin.site.unregister(GCMDevice)
    admin.site.register(GCMDevice, DeviceAdmin)

jamaalscarlett:
When you say NULL, do you mean that you saved a device with no device_id set?  That should not thrown an exception, what version are you running?

alfaro28:
Yes, the device_id was null on my mysql db, I'm running 1.3.1 with django 1.8.2

jamaalscarlett:
Try this branch, it should be fixed.  https://github.com/jamaalscarlett/django-push-notifications/tree/bug_device_id_null

ekonstantinidis:
I'm having exactly the same issue.. The error i'm getting:

    TypeError at /admin/push_notifications/gcmdevice/
    hex() argument can't be converted to hex
    Request Method:	GET
    Request URL:	http://192.168.1.79:60773/admin/push_notifications/gcmdevice/
    Django Version:	1.8.3
    Exception Type:	TypeError
    Exception Value:	 hex() argument can't be converted to hex

jamaalscarlett:
Have you tried the pull request code?

ekonstantinidis:
@jamaalscarlett I've just tried it and it works! It would be awesome if it was merged :+1:

jamaalscarlett:
I addressed one issue jleclanche had, so hopefully soon.

jleclanche:
Landed.

ekonstantinidis:
Amazing! Thanks @jleclanche for merging and for django-push-notifications! It is awesome! Is there an estimation for a new release?

jleclanche:
I've considered making a new release now but I'm giving it more time to see if more bugfixes can land.

tasn:
Hey, any plan for a release? It's been a while, and this bug is pretty annoying. :)

Thanks.

jleclanche:
Ill try to do a release after the current blocked pr lands.

tasn:
Thanks.

PR #193, opened 2015-07-25
Fixes issue #192, added a null check before attempting to convert

jamaalscarlett:
device_id value to hex when sqlite or mysql is configured as the
backend db.


jleclanche:
Just check for `and obj.device_id` please.

PR #215, opened 2015-09-09
Handle GCM canonical IDs

AlesDokshanin:


jamaalscarlett:
Looks good.  Just a thought, how would you handle the following situation.
1. I install my app and register for gcm I receive reg id "foo"
2. I uninstall the app.
3. Before a push notification is attempted, I re-install the app, and receive a new reg id "bar"
4. The server attempts to send a push notification to all devices attached to my user.
5. Your fix will update the first device with the latest reg id "bar", but now I have two instances of the same device in the table.
Should you be checking for an existing entry with the reg id and de-activating all but the most recent device registered?

AlesDokshanin:
@jamaalscarlett if I got everything right, according to [docs] (https://developers.google.com/cloud-messaging/http-server-ref) GCM server should return **NotRegistered** error when we will try to send push on 'foo' device for the next time and then the device will be automatically made inactive.

AlesDokshanin:
@jleclanche I added some comments with links to documentation.

jleclanche:
Thanks, code looks good at a glance. @jamaalscarlett what do you think?

AlesDokshanin:
@jleclanche fixed newline character.

jamaalscarlett:
@AlesDokshanin In the situation I described the "foo" and "bar" registration ids belong to the same device.  When I replicate the steps above using the test bulk message in django admin, I receive this response:
All messages were sent: {u'failure': 0, u'canonical_ids': 1, u'success': 2, u'multicast_id': 6727419149781304041, u'results': [{u'message_id': u'0:1441858935091806%3449593ff9fd7ecd'}, {u'registration_id': u'APA91bHP3uY62qkKZ5TsgSZjH386RsrErmIT3rSXXSNzXER5GoL0CA6QEG6lUx-mLND3Wn3SXSzVf2BU6Ry3la6z8mwroRdChqwVzKXht9WHCaNyrZK9PWCSUNj4Jrg99qNUAqbjzgXF', u'message_id': u'0:1441858935091278%3449593ff9fd7ecd'}]}
As you can see both registration ids return a result of successful, but the canonical_ids count is 1 because they are both the same device, but if I were to use the canonical id to update the registration id, I would then have two devices with the correct id.  This doesn't break anything, but I think deactivating one of the devices would be nice.  Other than that, the code looks great.

AlesDokshanin:
@jamaalscarlett thanks, you are right. I guess there were reasons not to set unique constraint to `registration_id` field and we also can't delete devices, but should deactivate them instead. Then I think I can add some check if there is another active device with given registration ID exists before assigning canonical ID. What do you think about it @jleclanche ?

AlesDokshanin:
@jleclanche could you please review what @jamaalscarlett said about deactivating all devices except one, if we are changing device ID to canonical. I'm ready to implement it, but it would be great to get your approval first.

AlesDokshanin:
@jamaalscarlett on the other hand, the example you gave is wrong example of using GCM service IMO, let me explain why: the most common case of using push notifications is when device is linked to user model in DB. And registration ID should be retrieved after login and removed after logout (or switching user). There shouldn't be more than 1 active GCMDevice records for one physical device, if there are - you're probably doing something wrong.

jamaalscarlett:
That is correct, but there is no way, from you app, to detect an uninstallation and then execute some code.  The only way we know that the app has been uninstalled, is when we attempt to send a notification and the gcm server responds NotRegistered.  The edge case I described involves uninstallation of the app, not a user logout.

AlesDokshanin:
@jamaalscarlett that makes sense. Then we should do something like this: if we're getting canonical ID when sending GCM message we should check if there are active devices with new canonical ID exist and then we've got 2 cases:
1. If device already exists, deactivate current device we got canonical ID for.
2. If there are no devices with canonical ID then replace current registration ID with canonical one.
Correct me if I'm wrong. Thanks for feedback anyway :)

jamaalscarlett:
That's correct.

jleclanche:
@AlesDokshanin Deferring to @jamaalscarlett's thoughts on this. I agree with the general logic.

AlesDokshanin:
@jleclanche @jamaalscarlett I updated the commit. Could you please check it out now?

jamaalscarlett:
Looks good to me.

AlesDokshanin:
@jleclanche could you please review the code?

AlesDokshanin:
@jleclanche bump

jleclanche:
Code still looks good. @jamaalscarlett your call

Issue #195, opened 2015-07-28
"No module named authorization" when using APNSDeviceResource and GCMDeviceResource with TastyPie 0.12.2


kenosan:
Hi, 

Trying to use django-push-notifications with Tastypie to expose end-points to allow registrations of devices. 

```python
from django.conf.urls import include, url
from push_notifications.api.tastypie import APNSDeviceResource, GCMDeviceResource
from tastypie.api import Api

v1_api = Api(api_name='v1')
v1_api.register(APNSDeviceResource())
v1_api.register(GCMDeviceResource())

urlpatterns = [
    url(r'^api/', include(v1_api.urls)),
]
```

When I try to hit http://localhost:8000/api/v1/?format=json, I get the following Traceback in the browser:

```
Environment:


Request Method: GET
Request URL: http://localhost:8000/api/v1/

Django Version: 1.8.3
Python Version: 2.7.6
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'mixtag',
 'tastypie',
 'push_notifications')
Installed Middleware:
('django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware')


Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  119.                 resolver_match = resolver.resolve(request.path_info)
File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py" in resolve
  366.             for pattern in self.url_patterns:
File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py" in url_patterns
  402.         patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py" in urlconf_module
  396.             self._urlconf_module = import_module(self.urlconf_name)
[...]

PR #225, opened 2015-09-27
Fixes module import issue for tastypie (issue #195)

kenosan:
Module import resolution seems to be broken because api/tastypie.py
imports tasypie items. Renaming api/tastypie.py to anything but
tastypie.py seems to resolve the import issue.

https://github.com/jleclanche/django-push-notifications/issues/195


jleclanche:
This is backwards incompatible and cannot possibly be the correct fix. We use local imports...

jleclanche:
Closing due to lack of activity.

PR #233, opened 2015-10-20
Correcting rest framework validation of GCMDevice.device_id max value.

lukeburden:
Half of the valid GCM device_id range currently elicits a `ValidationError` due to incorrect validation added in e8d07766e2e89b6ff176e386e88039a3c8cc1554, whereby the submitted value is checked against the signed 64 bit integer maximum rather than the unsigned.

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -2,7 +2,7 @@ from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext_lazy as _
 from .models import APNSDevice, GCMDevice, get_expired_tokens
-from django.db import connection
+
 
 User = get_user_model()
 
@@ -57,18 +57,5 @@ class DeviceAdmin(admin.ModelAdmin):
 			d.save()
 
 
-class GCMDeviceAdmin(DeviceAdmin):
-	"""
-	Inherits from DeviceAdmin to handle displaying gcm device as a hex value
-	"""
-	def device_id_hex(self, obj):
-		if connection.vendor in ("mysql", "sqlite") and obj.device_id:
-			return hex(obj.device_id).rstrip("L")
-		else:
-			return obj.device_id
-	device_id_hex.short_description = "Device ID"
-
-	list_display = ("__str__", "device_id_hex", "user", "active", "date_created")
-
 admin.site.register(APNSDevice, DeviceAdmin)
-admin.site.register(GCMDevice, GCMDeviceAdmin)
+admin.site.register(GCMDevice, DeviceAdmin)
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
 
-from django.db.backends.base.operations import BaseDatabaseOperations
 from rest_framework import permissions
 from rest_framework.serializers import ModelSerializer, ValidationError
 from rest_framework.validators import UniqueValidator
@@ -9,9 +8,7 @@ from rest_framework.fields import IntegerField
 
 from push_notifications.models import APNSDevice, GCMDevice
 from push_notifications.fields import hex_re
-
-
-BIGINT_MAX_VALUE = BaseDatabaseOperations.integer_field_ranges["BigIntegerField"][1]
+from push_notifications.fields import UNSIGNED_64BIT_INT_MAX_VALUE
 
 # Fields
 
@@ -22,7 +19,8 @@ class HexIntegerField(IntegerField):
 	"""
 
 	def to_internal_value(self, data):
-		# validate that value is a hex number
+		# validate hex string and convert it to the unsigned
+		# integer representation for internal use
 		try:
 			data = int(data, 16)
 		except ValueError:
@@ -78,9 +76,8 @@ class GCMDeviceSerializer(ModelSerializer):
 		}
 
 	def validate_device_id(self, value):
-		# max value for django.db.models.BigIntegerField is 9223372036854775807
-		# make sure the value is in valid range
-		if value > BIGINT_MAX_VALUE:
+		# device ids are 64 bit unsigned values
+		if value > UNSIGNED_64BIT_INT_MAX_VALUE:
 			raise ValidationError("ValidationError Device ID is out of range")
 		return value
 
--- a/push_notifications/fields.py
+++ b/push_notifications/fields.py
@@ -1,20 +1,45 @@
 import re
 import struct
 from django import forms
+from django.core.validators import MaxValueValidator
+from django.core.validators import MinValueValidator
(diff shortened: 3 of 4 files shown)


jamaalscarlett:
Have you verified this against postgresql?  I think i used the signed value because of a postgres limitation.

lukeburden:
I'm using postgres, yes.

A note on the second commit I've made here; `HexIntegerField.to_python` was breaking due to receiving integers larger than 2**63 due to my correction in the rest framework serializer (first commit). This method was performing two functions, the conversion of a signed integer from a postgres backend back into the unsigned form, as well as the conversion to a hex string. Reading up on it, the proper home for the special postgres logic is [Field.from_db_value](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.from_db_value), which is the clean reverse of `get_prep_value` as of django 1.8.

Moving the conversion into `from_db_value` allows `to_python` to always assume what it is receiving is the unsigned integer form, at which point it's only duty is to convert to a hex string.

jamaalscarlett:
I thought part of the reason for the hex integer field was that psql did not support an unsigned big int?

jleclanche:
@jamaalscarlett yes, correct.

lukeburden:
That is correct. The approach taken (which I haven't changed) is to store device_id as a signed big int if the backend is postgres. The custom field packs and unpacks the unsigned value in and out of that field. These improvements put the packing and unpacking in the correct place in the custom field, fixing a breakage I was experiencing in the `to_python` method.

jleclanche:
@lukeburden I didn't take that approach because that means the data becomes nonportable (migrating it to/from postgres will cause unexpected issues).

lukeburden:
@jleclanche can you explain what approach you mean? The approach I just described is what the code base already does .. I'm not changing the fundamentals of the implementation, just fixing some problems that existed for the use of the included rest_framework API classes plus incorrect handling of the device ID range.

jleclanche:
Ok, I'll do a closer review tomorrow hopefully if I get time. BTW you might want to look at CONTRIBUTING.md on commit messages styleguide ;)

jamaalscarlett:
@jleclanche as long as you are using manage.py dumpdata and loaddata, django should handle formatting the device_id properly.
@lukeburden you will also need to change fields.py.  HexIntegerField currently is a subclass of BigIntegerField which has a max value of 9223372036854775807 so anything higher would fail validation if the backend is mysql or sqlite.

jleclanche:
dumpdata/loaddata is very buggy, recommended solutions are usually [pgloader](http://pgloader.io/) and similar non-pg tools.

lukeburden:
@jleclanche, noted re. commit messages. I'll rebase and adjust.
@jamaalscarlett, I added a test for that scenario. See `test_rest_framework:test_device_id_validation_value_between_signed_unsigned_64b_int_maximums`. I see you guys use SQLite as your backend for tests and the test seems happy. This works as HexIntegerField has a custom `db_type` method such that on MySQL/SQLite the underlying column type will be unsigned big int.

jamaalscarlett:
@lukeburden your test is for the drf api.  I'm referring to Django admin.  Because the field is a BigIntegerField, when you attempt to save the device, the validators for BigIntegerField will be run against the value, and one validator checks the value against the max value for the field type which is 9223372036854775807

lukeburden:
@jamaalscarlett, I see what you mean. I've adjusted the validators there to restrict to the appropriate range. I've also changed it such that this validation is agnostic of the backend, as it's always performed against the unsigned value produced from the hex string. This means in the admin you get sensible validation error messages regardless of the backend you're using.

While I've got you - can you point me to some documentation on how to run the tests for this library? I can only run 25 of the full set of tests, not matter the command line args I use.

jamaalscarlett:
Using python tests/runtests.py, 31 tests are executed.  But when I execute tox, only 25 tests are executed.  I'm not sure why that is happening.

jamaalscarlett:
PR #235 will fix the issue with the tests.
With your latest change mysql admin is good, but I see this error when using sqlite: Python int too large to convert to SQLite INTEGER

lukeburden:
@jamaalscarlett, that PR results in 31 tests being run for me. I did find issues with a couple of tests, and I've fixed those. My tests are running against SQLite though and I'm not seeing the error you reference. Can you be more descriptive as to where it is happening?

jamaalscarlett:
@lukeburden I see the error when I attempt to save a GCM device with device_id 0x8000000000000000 using the django admin

lukeburden:
@jamaalscarlett, the problem you're experiencing there exists in master of this project for sqlite3, but we can use this pull request to resolve it. I had a chance to look into all this today, and learned some rather interesting things about sqlite3's strategy of storing integers. :/

HexIntegerField. db_type specifies a type of "UNSIGNED BIG INT" in the case the backend is sqlite3. This "type" actually translates to a more general integer type, which is always stored as a signed value. So the range for device_id in this project as it currently stands is -(2**63-1) to 2**63-1. This is why you were seeing the error you were, @jamaalscarlett.

Fortunately we can just enable the same handling logic for sqlite3 as I've submitted in this pull request as a way of dealing with postgres' similar situation.

For existing installations using sqlite device_ids will now be converted to unsigned values. You could use a conditional data migration to do this conversion for sqlite backends. Or just warn people and leave it it up to them. At the end of the day, device_id isn't critical for actually sending push notifications.

jleclanche:
I'm going to give a reserved gtm on this PR once it's been tested to work properly on all 3 engines. Please ensure there's no migration issue as well.

lukeburden:
@jleclanche - what's a gtm? I'm close to out of time to spend on this PR for now. Let me know if you have any further questions when you get a chance to have a proper look at it.

jamaalscarlett:
Good to merge.  I can pull it down and run it against all three this evening.  I'm on GMT -5 so wont be for 12 hours or so.

lukeburden:
Right, thanks gents. Also, for interest's sake here's a decent write-up on sqlite3 and its quirks when storing large integers:

http://jakegoulding.com/blog/2011/02/06/sqlite-64-bit-integers/

jamaalscarlett:
django admin works, and all tests pass when run against all three back ends.  Your changes mean that GCMDeviceAdmin is no longer nesseccary.  In admin.py, admin.site.register(GCMDevice, GCMDeviceAdmin) can be reverted to admin.site.register(GCMDevice, DeviceAdmin), as device_id_hex throws an error on your branch.

lukeburden:
Thanks for spotting that - I've removed the custom GCM admin class and updated the PR.

jamaalscarlett:
It's working great, nice job.  Any other thoughts @jleclanche ?

jleclanche:
You can land it if it works. Code looks good.

jamaalscarlett:
Landed.

PR #235, opened 2015-10-22
Added djangorestframework to tox deps list.

jamaalscarlett:
The tests in test_rest_framework were not running because they require djangorestframework.


jleclanche:
gtm

PR #245, opened 2015-11-30
Dont send to non-active devices

jschneier:


jleclanche:
@jamaalscarlett Can you review this please? I don't know if it's necessary.

jamaalscarlett:
I will take a look at this today.

jamaalscarlett:
So I have been thinking about this and I do not think that this is necessary.  More over I think that we have conflated active and registered.  If I have a mobile app, I may allow the user to turn off push notifications, thus the device would be inactive.  From Googles own dev page: **"For purposes such as alerting users, chat messaging or kicking off background processing before the user opens the client app, GCM provides a reliable and battery-efficient connection between your server and devices."**  So even if the user has decided to block push notifications, we may still need to ability to send messages to the app.  On the other hand if we attempt to send a message and the GCM server sends back an unregistered code, the mobile app is either uninstalled or re-register with another code.  The change I am proposing would require a new field and changes to the current bulk message method.  What do you think @jleclanche ?

jleclanche:
@jamaalscarlett Breaking changes are fine but not in this release. If you tell me this issue is not necessary then I won't include it in the next release.

jamaalscarlett:
Understood.  I would not include this.

PR #226, opened 2015-09-28
Added support for multiple APNS certificates

matthewh:
Supersedes PR #182.

This update introduces a new setting, APNS_APP_CERTIFICATES, and forcefully
requires developers to replace APNS_CERTIFICATE. Settings are
validated at startup using Django system checks. Based on the work of
@cristiano2lopes with minor updates to address outstanding issues
mentioned by the package maintainer.

Fixes: #130

PR #247, opened 2015-12-01

nnseva:

- old-style - one application per server instance
- new-style static - multiple (several) applications per server instance, new application requires settings patching or reloading
- new-style dynamic - multiple (a lot) applications per server instance, using custom model to provide an application key/certificate

This patch solves #130

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -173,7 +173,7 @@ reflect an application ID to the correspondent application credentials:
 
     class ApplicationModel(models.Model):
         application_id = models.CharField(max_length=64,db_index=True)
-        cgm_api_key = models.TextField()
+        gcm_api_key = models.TextField()
         apns_certificate = models.FileField()
 
 Then settings for the application should look like:
@@ -185,7 +185,7 @@ Then settings for the application should look like:
 		"GCM_API_KEYS_MODEL": {
 		    "model":"applications.ApplicationModel",
 		    "key":"application_id",
-		    "value":"cgm_api_key",
+		    "value":"gcm_api_key",
 		},
 		"APNS_CERTIFICATE": "/path/to/your/default/certificate.pem",
 		"APNS_CERTIFICATES_MODEL": {
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,8 @@ GCM or APNS devices and in the action dropdown, select "Send test message" or "S
 Note that sending a non-bulk test message to more than one device will just iterate over the devices and send multiple
 single messages.
 
+The application supports multiple mobile applications with separate application IDs using one server simultaneously.
+
 Dependencies
 ------------
 Django 1.8 is required. Support for older versions is available in the release 1.2.1.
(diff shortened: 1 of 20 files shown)


README.rst
```diff
@@ -117,6 +133,73 @@ Sending messages in bulk
...
+        apns_certificate = models.FileField(null=True,blank=True)
+
+Then settings for the application should look like:
+
+.. code-block:: python
+
+	PUSH_NOTIFICATIONS_SETTINGS = {
+		"GCM_API_KEY": "<your default application api key>",
+		"GCM_API_KEYS_MODEL": {
+		    "model":"applications.ApplicationModel",
```
jamaalscarlett: I'm pretty sure @jleclanche was against storing keys in the database

nnseva: It's a library user choice to have a key either in the dictionary, or in the database IMHO. Look to the code deeply, you will see a kind of model-stored dictionary which can be used on the place of the static dictionary transparently.

jleclanche: I'm still against keys in the db. You'll have to justify it a lot more thoroughly to change my mind about it. :)

nnseva: 1. As minimum two users of your library want it.
2. The productivity reason is not relevant - you will spend much more time to work with a lot of device registration records.
3. The security reason is not relevant - you can not restrict library users to store keys in the model and patch static dictionary using data from this model.
4. Using of model-stored keys may be an application need (it's my case). Independently of library originator means (sorry) the library user will implement some kind of settings dictionary patching or replacement (look to the patch code, push_notifications/modeldict.py file for example).
5. So, the most rational decision in this case - to follow user needs to make user life easier, not harder.

nnseva: And look to the patch, it allows the both, "static" dictionary and "dynamic" model-driven settings.

nnseva: @jleclanche you are answering to @evdoks in #130:
> Putting settings in the database prevents people from being able to manage their settings by file, simple as that.

My patch allows using old-style settings string, simple settings dictionary as simple as you want, and custom application model for extended users. All kinds of settings may be used together or separately and work well.

nnseva: @jleclanche you are answering to @t-io in #130:

> I have a very, very solid -2 on depending on db settings here.

You are not depending on database settings in my case. You still free to use single setting, or static dictionary.

Even when you are using model-driven settings, you don't not depend on a database structure, and don't restrict a user to have any database structure. You can use any model to have a storage for keys and certificates using my patch.

nnseva: Also @niktto wanted this patch I think

nnseva: Just read https://github.com/nnseva/django-push-notifications/blob/master/README.rst#multiple-mobile-applications to understand better what is implemented in a patch

niktto:
@nnseva YES! I needed this in 2 separate projects already. I owe you many, many beers :). Thank you. :+1:

t-io:
:+1: That looks very gooood :)

jschneier:
@nnseva I am using this branch now. You need to add 'application_id' to the rest framework serializer [here](https://github.com/nnseva/django-push-notifications/blob/master/push_notifications/api/rest_framework.py#L37).

nnseva:
@jschneier done

nnseva:
closed in regards to #259

nnseva:
Closed in regards to #260

PR #258, opened 2015-12-15
Removed call to with_metaclass as it is no longer needed

jamaalscarlett:
Subfieldbase is deprecated and scheduled to be removed with Django
1.10


jleclanche:
@jamaalscarlett Landed with a fixed commit message in 6988afd - please mind the commit message style please :)

PR #259, opened 2015-12-15

nnseva:
This is a replacement for the pull request #247 with rearranged and partially squashed commits.

- old-style - one mobile application per server instance, compatibility mode
- new-style static - multiple (several) mobile applications per server instance, settings file contains direct description of mobile applications credentials  (key/certificate), a new mobile application requires settings patching or reloading
- new-style dynamic - multiple (a lot) mobile applications per server instance, settings file contains reference to the custom model with mobile application credentials (key/certificate), a new mobile application can be created at runtime without server restart

This patch solves #130

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,8 @@ GCM or APNS devices and in the action dropdown, select "Send test message" or "S
 Note that sending a non-bulk test message to more than one device will just iterate over the devices and send multiple
 single messages.
 
+The application supports multiple mobile applications with separate application IDs using one server simultaneously.
+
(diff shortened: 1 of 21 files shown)


jleclanche:
You still have bad merges and rebases going in there... please clean it up, it can't be reviewed if the commits are not clean.

nnseva:
I'm afraid that I couldn't rearrange existent commits, just because several ones are mixed with other's. I can just make a single commit uniting jTeego and mine changes here, would it be enough?

Issue #148, opened 2015-05-08
Is it planned to support certificates with passphrase in future?


eyscode:
It'd be great if it's added support for certificates with passphrase for security reasons. If not, why? Thanks for your time.

jleclanche:
See #121 - you can't input them from django-push-notifications. I can't advise on the setup, though. Feel free to discuss it further in that issue.

Issue #121, opened 2015-01-08
Support passphrase for APNS


gorros:
Hi,

Are you planing to add support for APNS certificates with passphases?

Looking for reply,
Thanks.

jleclanche:
You're free to set a passphrase on one, but you'll have to enter it on the server itself I guess. I never really tested.

How would you see it work?

geno6:
I think you should use pass-phrase to decode and then open the certificate in python, passphrase can be kept anywhere.

jleclanche:
Clearly this is already the case, it's an interface problem.

gorros:
Yes, it works, it asks passphrase.

eriktelepovsky:
Ignoring passphrase existence is not a good solution by my opinion. Certificate secured by passphrase could be opened using OpenSSL library.

Like this:
```
def _connect(self, certificate, private_key, passphrase=None):
    """
    Establishes an encrypted SSL socket connection to the service.
    After connecting the socket can be written to or read from.
    """
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    cert = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, certificate)
    args = [OpenSSL.crypto.FILETYPE_PEM, private_key]
    if passphrase is not None:
        args.append(str(passphrase))
    try:
        pkey = OpenSSL.crypto.load_privatekey(*args)
    except OpenSSL.crypto.Error:
        raise InvalidPassPhrase
    context = OpenSSL.SSL.Context(OpenSSL.SSL.TLSv1_METHOD)
    context.use_certificate(cert)
    context.use_privatekey(pkey)
    self.connection = gevent_openssl.SSL.Connection(context, sock)
    self.connection.connect((self.hostname, self.PORT))
    self.connection.set_connect_state()
    self.connection.do_handshake()
```

PR #263, opened 2015-12-18
Certificate draft check for apns

nnseva:
This patch fixes a bad case when the server asks for the passphrase on the console when the certificate encrypted by passphrase is used, like #121 and #148 has been asked. Also it resolves issues like #2 passing well-defined description of the error in the exception.


jleclanche:
Looks mergeable as is, but you really should be doing this in a Django check:

https://docs.djangoproject.com/en/1.9/ref/checks/

Can you take care of this? I asked for it in another PR but it never materialized.

nnseva:
I didn't use this subsystem before yet, so I am not sure that I would implement it good enough. As I understand, it is appropriate for static checking before server execution, but is not a good choice for runtime checks of dynamic objects. Anyway, it would be better to implement features step-by-step.

PR #267, opened 2016-01-08
Clarify documentation around what APNS_HOST should be set as

rogueleaderr:
I wasted a few minutes being confused about why push notifications were being delivered when my site was running locally but not when I deployed it to a staging server. It looks like the non-sandbox APNS gateway will only work with app store builds. Setting the gateway to the sandbox when in staging (even with DEBUG=False) got notifications delivering again for me. Hopefully this will save someone else some time in the future.


rogueleaderr:
Well I feel silly but it appears this change is wrong. At first Heroku would only deliver notifications when the gateway was set to sandbox (or so it seemed), but now it'll only deliver them when it's not. 

I'll test some more and resubmit if my proposed solution turns out to be accurate.

Issue #268, opened 2016-01-10
Updating a device via PUT with Django-Rest-Framework does not update the Device.user


salian:
My workflow using Django-Rest-Framework is such: 
1) Register a device when my Android app is run for the first time. I am following a lazy registration pattern, so at this point, the user is not logged-in, and I have to use the `GCMDeviceViewSet` instead of the `GCMDeviceAuthorizedViewSet`.
2) When the user logs in at some future date, I wish to update the registered device with the user parameter. For this, I send a PUT request at the `/device/gcm/{registration_id}` detail endpoint.

Question: Does this workflow seem appropriate?

Issue: At this point, although the appropriate Authentication token is passed in the header (i.e. the user is logged-in) the `Device.user` is not updated. I think this is a bug, due to `perform_update()` not being implemented in `DeviceViewSetMixin()` in `rest_framework.py`.

My suggested fix is to override `perform_update()` in the `DeviceViewSetMixin` in `rest_framework.py`, similar to the existing `perform_create`

	def perform_update(self, serializer):
		if self.request.user.is_authenticated():
			serializer.save(user=self.request.user)
		return super(DeviceViewSetMixin, self).perform_update(serializer)

jamaalscarlett:
@salian Why do you need to register the device as soon as the app is launched.  You really should only be registering the device after the user has registered and logged in.

salian:
@jamaalscarlett Almost half of my app is usable without logging in. I want to retain the ability to send the occasional generic notification about fresh content to those users, aimed at getting them to use the app more.

jamaalscarlett:
Fixed in #269

PR #269, opened 2016-01-10
Update Device.user when updating existing devices via DRF.

salian:
Addresses issue #268 

Overrides `perform_update` (similar to `perform_create`) to update Device.user when updating existing devices.

### DIFF
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -103,7 +103,7 @@ class DeviceViewSetMixin(object):
 			serializer.save(user=self.request.user)
 		return super(DeviceViewSetMixin, self).perform_update(serializer)
 
-	
+
 class AuthorizedMixin(object):
 	permission_classes = (permissions.IsAuthenticated, IsOwner)
 
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -98,7 +98,12 @@ class DeviceViewSetMixin(object):
 			serializer.save(user=self.request.user)
 		return super(DeviceViewSetMixin, self).perform_create(serializer)
 
-
+	def perform_update(self, serializer):
+		if self.request.user.is_authenticated():
+			serializer.save(user=self.request.user)
+		return super(DeviceViewSetMixin, self).perform_update(serializer)
+		
+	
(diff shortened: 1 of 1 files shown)


jamaalscarlett:
This looks fine, although there is an inherent flaw in the un-authenticated viewset.  Currently, anyone can edit any object.  All that is needed is the registration_id, once that is know any of the serializer fields can modified.

salian:
@jamaalscarlett I agree, especially in combination with an un-authenticated endpoint for exposing a list of all devices via GET. It takes away whatever little security via obscurity there is. 
I suppose I shall want to exclude the `RetrieveModelMixin` from the un-authenticated view `GCMDeviceViewSet`so it doesn't give out a list of all my registration IDs.

jamaalscarlett:
@salian I would be more comfortable if there was a check for an existing user.  To prevent the current user for being overwritten, I would only save the user if the current user is None.

salian:
I think the desired behaviour is to overwrite the `user` in case of a duplicate `registration_id`. 
As I understand, there is explicitly a unique validator on the `registration_id` field at `line 71` in `rest_framework.py`, so having two records with the same `registration_id` is not an option.

Moreover GCM will issue the same registration id again unless a user uninstalls and reinstalls the app, so there is no way for the calling app to come up with a different registration_id; in which case arguably it is better to keep the fresh record than the stale one?

jamaalscarlett:
@salian I forgot about the validator.  This is why a wrote my own DRF integration.  Could you please squash these commits.  After that, this is good to merge.

jamaalscarlett:
@salian can you clean up the commit so we can get this merged in?

jamaalscarlett:
@jleclanche any thoughts on this one?

salian:
@jamaalscarlett I actually can't figure how to clean it via the web interface, and I am not familiar with the CLI. Should I delete this and make a fresh branch?

jamaalscarlett:
From the command line run git rebase -i HEAD~3 and follow the instructions.

jamaalscarlett:
@salian any luck?  I can fix it if you still can't figure it out.

jleclanche:
@jamaalscarlett gtm; you can rebase & land. sorry this went under the radar.

PR #273, opened 2016-01-21
Fix bug when using gcm serializer as a nested serializer.

### LINKED ISSUE #270: registration_id unique constraint fails on PUT (Update)
We use GCMDeviceSerializer as nested serializer
```python
class DataSerializer(serializers.ModelSerializer):
    gcm_device = GCMDeviceSerializer(allow_null=True, required=False)
```
and it always fails validation when we try to update our Data object with message: `{"gcm_device":{"registration_id":["This field must be unique."]}}`

It seems that following code executes on both create and update calls:
```python
			# Work around an issue with validating the uniqueness of
			# registration ids of up to 4k
			'registration_id': {
				'validators': [
					UniqueValidator(queryset=GCMDevice.objects.all())
				]
			}
```

jamaalscarlett:
When the serializer is nested the unique validator on registration_id
does not behave correctly.  This is a known issue with DRF
see: https://github.com/tomchristie/django-rest-framework/issues/2403

Added missing allow_null to device_id field

Fixes #270

### DIFF
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 
 from rest_framework import permissions
 from rest_framework.serializers import ModelSerializer, ValidationError
-from rest_framework.validators import UniqueValidator
 from rest_framework.viewsets import ModelViewSet
 from rest_framework.fields import IntegerField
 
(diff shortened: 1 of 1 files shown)


jleclanche:
gtm but please fix the issue ref from `jleclanche#270` to `#270` in the message first.

PR #278, opened 2016-02-03
Enabled support for arbitrary user model

fsto:
This enables the use of custom models.

A user can define which model to use by setting the `PUSH_NOTIFICATIONS_SETTINGS` key `USER_MODEL` to a model on the format `appname.Model`. The default user model is `settings.AUTH_USER_MODEL` which is the current user model in the library.

The `admin.py`'s `search_fields` only contains the user model's `USERNAME_FIELD` if the user model actually has that field.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,6 +75,7 @@ For APNS, you are required to include ``APNS_CERTIFICATE``.
 - ``APNS_PORT``: The port used along with APNS_HOST. Defaults to 2195.
 - ``GCM_POST_URL``: The full url that GCM notifications will be POSTed to. Defaults to https://android.googleapis.com/gcm/send.
 - ``GCM_MAX_RECIPIENTS``: The maximum amount of recipients that can be contained per bulk message. If the ``registration_ids`` list is larger than that number, multiple bulk messages will be sent. Defaults to 1000 (the maximum amount supported by GCM).
+- ``USER_MODEL``: Your user model of choice. Eg. ``myapp.User``. Defaults to ``settings.AUTH_USER_MODEL``.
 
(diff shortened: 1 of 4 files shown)


jleclanche:
Could you fixup the second commit into the first one please?

fsto:
@jleclanche of course, this is now fixed.

jleclanche:
lgtm but can you explain how this is necessary? why can't we just use the user model from the setting?

fsto:
@jleclanche ok! There are cases when you have multiple user models.

A current case for me is that I've integrated with another system which holds the users I want to send push notfications to. The users (via their mobile clients) communicate with one aggregating server which communicates with (1) an external server holding user data and (2) the django server which complements the external server with extra user data and push notification functionality. In this case it therefore makes sense to send push notifications to another user model than `AUTH_USER_MODEL`.

jleclanche:
Ok. I'm not too happy with the solution but I think it's fine to expose that field anyway.

jleclanche:
Remove the try block and I'll land it.

fsto:
@jackfengji just tell me how you'd be happy with the solution.

jleclanche:
I meant more in a general sense that I don't think the correct way to solve your issue is to change that. But I am fine with exposing that as a setting anyway, so might as well.

fsto:
Fixed @jackfengji

jleclanche:
Landed in 40cfe2b

rromanchuk:
@jleclanche this is a nice edition, some people don't extend any form of the django user object but still wan't these relationships. Thx

PR #282, opened 2016-02-25
#290 do not try to send any messages if registration_ids is empty

gitfred:
also test added

### DIFF
--- a/push_notifications/models.py
+++ b/push_notifications/models.py
@@ -10,91 +10,93 @@ from .settings import PUSH_NOTIFICATIONS_SETTINGS as SETTINGS
 
 @python_2_unicode_compatible
 class Device(models.Model):
-	name = models.CharField(max_length=255, verbose_name=_("Name"), blank=True, null=True)
-	active = models.BooleanField(verbose_name=_("Is active"), default=True,
-		help_text=_("Inactive devices will not be sent notifications"))
-	user = models.ForeignKey(SETTINGS["USER_MODEL"], blank=True, null=True)
-	date_created = models.DateTimeField(verbose_name=_("Creation date"), auto_now_add=True, null=True)
+    name = models.CharField(max_length=255, verbose_name=_("Name"), blank=True, null=True)
+    active = models.BooleanField(verbose_name=_("Is active"), default=True,
+        help_text=_("Inactive devices will not be sent notifications"))
+    user = models.ForeignKey(SETTINGS["USER_MODEL"], blank=True, null=True)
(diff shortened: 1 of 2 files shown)


push_notifications/models.py
```diff
@@ -10,91 +10,93 @@
...
-			if message is not None:
-				data["message"] = message
+            data = kwargs.pop("extra", {})
+            if message is not None:
+                data["message"] = message
 
-			reg_ids = list(self.filter(active=True).values_list('registration_id', flat=True))
-			return gcm_send_bulk_message(registration_ids=reg_ids, data=data, **kwargs)
+            reg_ids = list(self.filter(active=True).values_list('registration_id', flat=True))
+            if not reg_ids:
```
gitfred: here comes the change

push_notifications/models.py
```diff
@@ -10,91 +10,93 @@
 
 @python_2_unicode_compatible
 class Device(models.Model):
-	name = models.CharField(max_length=255, verbose_name=_("Name"), blank=True, null=True)
-	active = models.BooleanField(verbose_name=_("Is active"), default=True,
-		help_text=_("Inactive devices will not be sent notifications"))
-	user = models.ForeignKey(SETTINGS["USER_MODEL"], blank=True, null=True)
-	date_created = models.DateTimeField(verbose_name=_("Creation date"), auto_now_add=True, null=True)
+    name = models.CharField(max_length=255, verbose_name=_("Name"), blank=True, null=True)
```
jamaalscarlett: This project uses tabs, not spaces.  Please see: [CONTRIBUTING.md](https://github.com/jleclanche/django-push-notifications/blob/master/CONTRIBUTING.md)

push_notifications/models.py
```diff
@@ -10,91 +10,93 @@
...
+            data = kwargs.pop("extra", {})
+            if message is not None:
+                data["message"] = message
 
-			reg_ids = list(self.filter(active=True).values_list('registration_id', flat=True))
-			return gcm_send_bulk_message(registration_ids=reg_ids, data=data, **kwargs)
+            reg_ids = list(self.filter(active=True).values_list('registration_id', flat=True))
+            if not reg_ids:
+                return
+            return gcm_send_bulk_message(registration_ids=reg_ids, data=data, **kwargs)
```
jamaalscarlett: Wouldn't it be easier to put the check in the gcm_send_bulk_message method?  That would remove the need to check for None every time you want to call the method.

jamaalscarlett:
@jleclanche Do you have a problem with me taking a shot at this one?  I think the initial issue is valid, a return would be better than a 400 error

jleclanche:
Go for it.

PR #288, opened 2016-03-28
new settings `POST_RECREATE_USER`

1vank1n:
Add new settings **POST_RECREATE_USER**.

This is necessary when user want reauthorized on one device in another account. Device `registration_id + user` will deleted and created with new authorization.

By default this property is disabled. For activation you need add to your `settings.py`:

```python
PUSH_NOTIFICATION_SETTINGS = {
   ...
   "POST_RECREATE_USER": True,
   ...
}
```

### DIFF
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+from django.shortcuts import get_object_or_404
+
 from rest_framework import permissions
 from rest_framework.serializers import ModelSerializer, ValidationError
 from rest_framework.viewsets import ModelViewSet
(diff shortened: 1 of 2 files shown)


jleclanche:
This doesn't seem like the sort of behaviour you would want in a setting. Why not handle it at the app level?

1vank1n:
@jleclanche, hi!

First, thank you for this cool project.
Second, How to handle it at the app level? Middleware? I think it's not obvious way.

Of course you can don't merge it. This fix helpful for me and I just want to share it. Maybe anybody in searching this solution now )

jleclanche:
If django doesn't offer a way to easily override that behaviour, you can always monkeypatch it. But so far I'm -1 on merging this.

jamaalscarlett:
I agree with @jleclanche this seems like behavior that should be handled at the app level.  -1 for me as well.

PR #291, opened 2016-04-23
gcm_send_message and gcm_send_bulk_message will no longer attempt to

jamaalscarlett:
send a gcm message if the registration id is None


jleclanche:
GTM but please fix the commit message

PR #299, opened 2016-06-02
Two little features

goinnn:
1. Add DEBUG like attribute of PUSH_NOTIFICATIONS_SETTINGS (very very simple and backward compatible)
2. Now push_notification allows push to several apps (simple and backward compatible)


jleclanche:
I'm a little confused about 08cba8e. What does it enable the library to do?

goinnn:
@jleclanche [08cba8e] this is for if you use your django server for two, three or more applications. For example in my case it is something like Uber (taxis and passengers)... I need two certificates for ios.

So now I can do something like this:

APNSDevice.objects.filter(pk__in=device_pks).send_message(....., certfile=TAXI_CERT)

or

APNSDevice.objects.filter(pk__in=device_pks).send_message(....., certfile=PASSENGER_CERT)

goinnn:
Another detail :-)

jleclanche:
Ok, I see.

I won't land the `DEBUG` commit, but I'll land the other two.

jleclanche:
Landed in f08d6aaf5285890c46742da96fb6175f0cd1ad2d

Issue #289, opened 2016-03-30
Windows Phone support


SvetlanaM:
Hi, do you consider Windows Phone push notification support? If no, I can create one and send pull request.

jleclanche:
PRs welcome yes.

gabrielpcosta18:
Are you already developing a support for Windows Phone?

jleclanche:
Nobody is working on this to my knowledge, feel free.

SvetlanaM:
I try to code a bit the next month.

gabrielpcosta18:
Really cool. This support would be pretty nice.

AdamBolfik:
Started working on WNS:

jleclanche/django-push-notifications#303

jamaalscarlett:
See https://github.com/jleclanche/django-push-notifications/commit/f32d9d50d48802954a34ee9e48c930ba556fab54

PR #303, opened 2016-06-15
WNS(Windows Notification Service) support

AdamBolfik:
I added the ability to send toast/tile/badge/raw notifications via WNS (Windows Notification Service). 

I have tested toast and tile notifications, but not badge or raw. I know badge should JUST work because it follows the toast and tile format closely. 

As for raw notifications: We are developing our app with ionic/cordova so I haven't even begun to see what raw notifications are even for. The docs say it's just a string passed to the app in the background so I made the methods handle this scenario. But again, I'm just assuming it works/works this way.

There is also some functionality that I haven't added, that would be nice:
* It looks as though WNS can get active devices back with the correct header passed <- use for pruning
* Each call to a create a push notification calls WNS for authentication <- This should be saved until expiration (included in response) to minimize calls.

I've added a little on the README, but it's not even close to complete. Thought that I would send this up for review as it is already massive.


push_notifications/admin.py
```diff
@@ -2,18 +2,17 @@
...
 User = apps.get_model(*SETTINGS["USER_MODEL"].split("."))
 
 
 class DeviceAdmin(admin.ModelAdmin):
 	list_display = ("__str__", "device_id", "user", "active", "date_created")
-	list_filter = ("active", )
+	list_filter = ("active",)
 	actions = ("send_message", "send_bulk_message", "prune_devices", "enable", "disable")
-	raw_id_fields = ("user", )
+	raw_id_fields = ("user",)
```
jleclanche: You're making unnecessary whitespace changes all over the place - these pollute the diffs and make it very hard to review the PR. Please don't do that.

push_notifications/wns.py
```diff
@@ -0,0 +1,359 @@
...
+		elif err.code == 406:
+			raise WNSNotificationResponseError("HTTP 406: The cloud service exceeded its throttle limit.")
+		elif err.code == 410:
+			raise WNSNotificationResponseError("HTTP 410: The channel expired.")
+		elif err.code == 413:
+			raise WNSNotificationResponseError("HTTP 413: The notification payload exceeds the 5000 byte size limit.")
+		elif err.code == 500:
+			raise WNSNotificationResponseError("HTTP 500: An internal failure caused notification delivery to fail.")
+		elif err.code == 503:
+			raise WNSNotificationResponseError("HTTP 503: The server is currently unavailable.")
```
jleclanche: There is really no need for this. You can `raise WNSNotificationResponseError(err)` instead.

push_notifications/admin.py
AdamBolfik: Sorry,  my ide went crazy on it.

push_notifications/wns.py
```diff
@@ -0,0 +1,359 @@
...
+												'children': 'text2',
+											},
+										],
+									},
+								},
+							},
+						},
+					},
+				},
+			}
```
jleclanche: Please don't format that dict like that, it's unreadable. Same comment applies to the tests.

push_notifications/wns.py
AdamBolfik: Yeah,  figured I would just make a more specific message. I mapped the errors to the wns docs.

push_notifications/wns.py
AdamBolfik: K, my ide didn't show sick large indents,  I'll look into it.

AdamBolfik: Such*

jleclanche:
Overall I like the code. The commit log can't stay like that though (see [CONTRIBUTING.md](https://github.com/jleclanche/django-push-notifications/blob/master/CONTRIBUTING.md)). If you don't know how to rebase it, I can do it.

I'll see if someone else can review/test it as well.

push_notifications/wns.py
AdamBolfik: Alright, I've looked into it... I'm not sure how to format this correctly with tabs. What's your preference? Just put it all one line-ish?

AdamBolfik:
Sorry, I noticed your contribution log after I started contributing...sigh.

If rebasing is something you do with ease, I'll gladly let you handle it. If not, I can take a look at it tomorrow.

jleclanche:
Well, it's a good skill to know regardless, so you can try:

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

You don't need to create a new PR. Force-pushing will update this one. If you get stuck/mess up completely, `git reset --hard 6cfcd32` will get you back to where you started :)

AdamBolfik:
Alright, I've rebased it. I just squashed the last couple commits about tabs/spaces as my other commits all seem to adhere to 'one change', at least, in my view.

jamaalscarlett:
It looks like the whitespace changes are still there.

AdamBolfik:
The whitespace keeps you on your toes. ;)

jamaalscarlett:
I've got two kids, I'm always on my toes.

push_notifications/wns.py
```diff
@@ -0,0 +1,359 @@
...
+except ImportError:
+	# Python 2 support
+	from urllib2 import HTTPError, Request, urlopen
+
+from django.core.exceptions import ImproperlyConfigured
+
+from . import NotificationError
+from .settings import PUSH_NOTIFICATIONS_SETTINGS as SETTINGS
+
+WNS_ACCESS_URL = "https://login.live.com/accesstoken.srf"
```
jleclanche: This should be a setting (cf. gcm/apns)

AdamBolfik: I'm fine with this either way, but what is the reason for making this configurable at the project level? 
It would seem that if WNS changes the url or api in any way it would have to be maintained by the app(you) and wouldn't be something the project developer would want/need to have knowledge about. 

Basically, If it needs to be overridden, shouldn't it be a bug that needs fixing in master?

AdamBolfik:
Also, this links back to #289 whenever this PR gets closed. Not sure if I needed to link it both ways.

push_notifications/wns.py
jleclanche: It's useful if the upstream URL changes and the DPN project is unmaintained at that time, and/or someone runs a self-hosted WNS dispatch server at a different address.

Both are unlikely, but it mirrors the behaviour in GCM/APNS.

push_notifications/wns.py
```diff
@@ -0,0 +1,359 @@
+"""
+Windows Notification Service
+
+Documentation is available on the Windows Dev Center:
+https://msdn.microsoft.com/en-us/windows/uwp/controls-and-patterns/
```
jleclanche: This URL is 404. What's the correct one?

jleclanche:
I'm going to touch up the commits and finish off the rebase locally but I would really like at least 1-2 people to tell me they tried it and it works fine.

I notice there's no migration for the WNSDevice model at the very least so I suspect it's not quite complete yet.

push_notifications/migrations/0003_wnsdevice.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
AdamBolfik: Here is the migration

jleclanche:
Anyone testing it?

jamaalscarlett:
@jleclanche I think it is going to be hard to find people both using this library and using windows phone.

jleclanche:
@jamaalscarlett Could you just run it to make sure it doesn't break existing installations? I'm mostly concerned about that.

jamaalscarlett:
Will do.  I'm @ work right now.  But I will run it when I get home tonight.

jleclanche:
@jamaalscarlett ping

jamaalscarlett:
The branch does not through any errors.

jleclanche:
Squashed, cleaned up and landed.

PR #315, opened 2016-08-28
Added FCM Support

AliLozano:
Still missing documentation, but it's ok to start.


push_notifications/fcm.py
```diff
@@ -0,0 +1,164 @@
...
+			raise FCMError(response)
+	return response
+
+
+def _fcm_handle_canonical_id(canonical_id, current_id):
+	"""
+Handle situation when FCM server response contains canonical ID
+"""
+	if FCMDevice.objects.filter(registration_id=canonical_id, active=True).exists():
+		FCMDevice.objects.filter(registration_id=current_id).update(active=False)
```
jleclanche: This doesn't look right.

AliLozano:
I said, its good to start, I didn't find how to send a bulk message with FCM.

jleclanche:
Yeah, I'm leaving the PR open for other people to build upon, but it's not mergeable as-is.

jamaalscarlett:
@AliLozano This looks like a lot of duplicate code.  Couldn't most of it be removed by a new column on the GCM model?  Something like isFCM?

AliLozano:
Yes, It could be a variable in a settings, but i think gcm will be removed in the future.

jamaalscarlett:
@AliLozano  Is this really needed?  This is from the fcm [faq](https://developers.google.com/cloud-messaging/faq) 

How is FCM server implementation different from GCM server implementation?

You won’t need to make any server-side protocol changes for the upgrade. The service protocol has not changed. However, note that all new server enhancements will be documented in FCM server documentation.

Have you tried just changing the url to point to the fcm endpoint and seeing if it works?

alexandreferreira:
Hi I tried an another approach for add FCM in #336 
thanks

PR #316, opened 2016-09-06
Firefox support

goinnn:
I'm sorry I know my PR is not very clean. In this PR you have firefox support.

I am going to do (in autumn) also Safari support and Edge  support. Chrome is supported with GCM.

For browser you need a simple model to save the payload info, I have this model in another app, but I can move it.

Please tell me if you are interested in these supported


jleclanche:
Thanks. So this is webpush, no sense making it Firefox-specific imo. Just make sure that the url is swappable.

The PR will have to be cleaned up to be merged, though - as it stands it's a fork pr.

jamaalscarlett:
@goinnn What is this actually doing?  So far I don't see where a message is sent.  And I don't see any of the encryption that webpush requires.

goinnn:
> Thanks. So this is webpush, no sense making it Firefox-specific imo. Just make sure that the url is swappable.

@jleclanche Every browser is very different. This implementation does not work for another browser, no sense swap the URL, although this url is set in the settings, for cleaning: https://github.com/goinnn/django-push-notifications/blob/develop/push_notifications/settings.py#L19

1. Chrome is GCM, currently works on my app (web, iOS and Android with ionic)
2. Firefox is very simple, it works from today. Only a post for an URL: https://github.com/goinnn/django-push-notifications/blob/develop/push_notifications/firefox.py#L8
3. Safari is the hell: https://developer.apple.com/library/mac/documentation/NetworkingInternet/Conceptual/NotificationProgrammingGuideForWebsites/PushNotifications/PushNotifications.html But a mate have implemented the frontend, and when I have any time I will implement the backend.
4. And Edge.... Edge is Internet Explorer with a costume: https://blogs.windows.com/msedgedev/2016/05/16/web-notifications-microsoft-edge/

> What is this actually doing? So far I don't see where a message is sent. And I don't see any of the encryption that webpush requires.

@jamaalscarlett The push on browser are without message... yes it is incredible!! The push system, e.g.: GCM  sends a notification for the browser and the browser requests the message to the backend (payload view)...  I have implemented this view and the model where the messages are saved ... So this if you are interested  in webpush I can prepare a nice PR.

jamaalscarlett:
@goinnn I'm still not sure I understand the purpose of https://github.com/jleclanche/django-push-notifications/pull/316/files#diff-2360c5d962b817cde3cabf9dae7c7547R6?  webpush does allow you to send messages:
https://github.com/chrisdavidmills/push-api-demo

goinnn:
@jamaalscarlett I have been studing the firefox push and I think firefox is currently the only browser that accept payload (messages): http://stackoverflow.com/questions/35950390/firefox-push-notifications-send-data-via-curl

Am I right?

jamaalscarlett:
@goinnn firefox is the only browser that uses the mozilla webpush spec.  Chrome uses GCM and Safari uses APNS.  I'm not sure what edge/internet explorer use.

alexbarcelo:
@goinnn If you use `endpoint` in an agnostic way and ensure to use the `keys` as intended (both `auth` and `p256dh`), you can send data from Django to both Firefox and Chrome. At least, that is what `pywebpush` does, and it does it well (at a much lower level than this, ofc). 

To send data to chrome you simply have to ensure to use the `gcm_key` when needed and the corresponding endpoint.

jamaalscarlett:
There is no actual functionality in this pull request.

PR #319, opened 2016-09-14
authentication token

null-none:
View set authentication token


jleclanche:
I won't be merging this for two reasons:

1. The token auth in DRF is actually very poor and often overridden by different token auth systems
2. Without any documentation or anything, it's just noise.

PR #321, opened 2016-09-20
Fix support for sending GCM messages to topic subscribers.

jamaalscarlett:
gcm_send_bulk_message did not allow passing a registration_ids value of None.
Update README.rst with correct example for send messages to topics

### DIFF
--- a/push_notifications/gcm.py
+++ b/push_notifications/gcm.py
@@ -107,7 +107,7 @@ def _gcm_send_json(registration_ids, data, **kwargs):
 	This will send the notification as json data.
 	"""
 
-	values = {"registration_ids": registration_ids}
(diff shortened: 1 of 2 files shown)


push_notifications/gcm.py
```diff
@@ -120,7 +120,7 @@ def _gcm_send_json(registration_ids, data, **kwargs):
 	data = json.dumps(values, separators=(",", ":"), sort_keys=True).encode("utf-8")
 
 	response = json.loads(_gcm_send(data, "application/json"))
-	if response["failure"] or response["canonical_ids"]:
+	if response.get("failure", None) or response.get("canonical_ids", None):
```
jleclanche: The second argument is not necessary, `.get()` will return `None` if you do not give it a second argument.

PR #324, opened 2016-09-30
Merge nneva multiple application changes

mlindemu:
Attempting to resolve conflicts between nneva's multiple application branch and jleclanche.


mlindemu:
Ok, I'm stuck here.  How do I merge migrations from two different commits when the library is only an app?

jleclanche:
There's a lot of stuff here. I can't tell what any of it does.

singleton11:
I agreed with @jleclanche, it should be divided to small features and every feature should splitted to many small commtis to make CR more easier.
Also I think, applications should be instances of some model, because I prefer to have an ability to manage applications from admin. And if this approach will be used, we have to store APNS Certificates into database, because some projects uses S3 backend for static and user media files, and there is a bad idea make certificates public (which contains private keys)
Therefore, openssl should be replaced by pyopenssl library to implement possibility to connect to APNS server with certificate from string. But more better is implementing this ability in PyAPNS and use it as external dependency. But this is another issue

jleclanche:
I definitely would like to offload the APNS logic to PyAPNS or a similar library. Same for GCM etc.

PR #327, opened 2016-10-04
Fix UniqueRegistrationSerializerMixin validate method.

jamaalscarlett:
If the request method is PATCH attrs is not guaranteed to contain
either "id" or "registration_id"


jamaalscarlett:
@jleclanche ?

PR #328, opened 2016-10-07
Add configuration option to make APNSDevice.registration_id unique_together with user_id, fixes #41.

### LINKED ISSUE #41: Removing the unique device constraint
I want to remove the unique device id constraint - there are a lot of situations where we want to allow multiple users to register on the same device.  What's the best way to do this?  Also does removing the constraint on the db level (/ model level? ) cause any parts of the code to change in its function?

eLod:
I have added the configuration option, but unsure about what to test (and how). Tests pass (except for py34 which i don't have local) with original settings and when adding the flag.


jleclanche:
@eLod Sorry if I was unclear; what I meant was that I would accept a PR to make it unique, not a unique_together. The unique_together approach doesn't make much sense as a generic approach imho.

eLod:
@jleclanche make unique what?

jleclanche:
The `registration_id`; like you did, but as `unique` instead of `unique_together`.

eLod:
@jleclanche that is already unique, i don't understand. the problem is it is unique by itself.

jleclanche:
@eLod Moving the unique to a setting would allow users to disable the uniqueness.

eLod:
@jleclanche so i should make it non-unique. would you accept it if it would allow 3 options, unique (default), non-unique and unique_together?

jleclanche:
I'm not comfortable with a tri-state for this. Adds too much complexity.

eLod:
Any way to accept unique_together? Like 2 configuration options?

jleclanche:
Why do you want it to be unique_together?

You are free to manually add the constraint to your database, Django won't complain.

eLod:
We are using the drf views, i rely on it returning an error instead throwing an exception.

jleclanche:
Can't you subclass the DRF views?

eLod:
I can, but it still wouldn't allow me to have the constraint in the model meta, which i consider the semantically correct solution.

jleclanche:
@jamaalscarlett Can I get your thoughts on this please?

eLod:
As i said in the issue #41 i see a valid use-case where a single device (and the same registration id) could be used for multiple accounts, but i still don't want to let the iOS app be able to register the same registration id for the same user account twice. I could handle all that with custom drf views and custom serializers, but i really consider the unique_together the simplest and semantically more correct solution (as in it reflects the reality better and even is simpler to implement and maintain).

jamaalscarlett:
@eLod @jleclanche I personally don't like the idea of a bunch of duplicate entries in the device table.  Why not handle it in your own django app?  Create a model that can map multiple users to the same entry in the device table.

KartikeyaRokde:
Hi @eLod : We are also running through exactly the same issue, can you please give us some light on which approach did you use to get your app running?
  - Did dropping the unique constraint in the database work?

rocchidavide:
We have also the same request

eLod:
@KartikeyaRokde sorry for the late response, what i did was actually moving the unique constraint to be on (user_id, reg. id) (e.g. on multiple columns). for that i had to modify a couple things.

PR #331, opened 2016-10-20
fix incompatibility with android device ids

macolo:
this fixes the `bigint out of range` error


jleclanche:
Like I said, it's not as easy as that. There are a ton of users of the library *today* who would be affected by this and have to run data migrations, which is too much to ask at this point.

macolo:
ok, we could include a custom migration?

jleclanche:
I'm extremely wary of data migrations in libraries. For example, if someone runs them on a very large table without paying much attention, they could lock their prod db down for a very long time.

The safest way to do it is to make this opt-in somehow, and include a management command to migrate the data. Once it's been opt in for one release or two, make a new major release that *only* makes it default behaviour.

jleclanche:
Ok, I thought about it some more and I think I found a safe way to do it without a setting that wouldn't lock the table too long:

1. Rename the hexint field to `device_id_legacy`, adding a migration for it
2. Add a *new* field `device_id`, a CharField which works much the same as the HexIntegerField (nullable, indexed).
3. Ensure that `device_id` works the same way as before in the DRF API. This is important, otherwise it breaks backwards compatibility in the API itself.
4. Ensure `device_id` gets filled properly, and `device_id_legacy` no longer does.
5. Add a data migration which casts all non-null `device_id_legacy` to their `device_id`.
6. Drop the `device_id_legacy` column
7. Remove `api.rest_framework.HexIntegerField`. The other HexIntegerField class will have to stay as it's in existing migration.

jleclanche:
I think @matthewh is handling the hexint migration stuff, so I'm closing this.

matthewh:
Yes I have a PR almost ready for this.

PR #334, opened 2016-10-25
Add possibility to use different badge values in bulk send

alexole:
It's very valuable to set a proper badge value for different users (for now only one value for all the pushes possible) even when sending messages in bulk, this is an example of usage:

APNSDevice.objects.filter(user_id__in=user_ids).send_message(
        text, 
        badge=lambda token: APNSDevice.objects.get(registration_id=token).user.get_badges()
)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,8 @@ value per user. Assuming User model has a method get_badge returning badge count
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


jleclanche:
Thank you, looks good. Can you add some documentation about it in the readme?

alexole:
Sure, will add into README.

alexole:
Is that ok?

jleclanche:
Landed in 0f90e14.

Issue #296, opened 2016-05-12
Changing User on a Device


chozabu:
Heyah - I see we can DELETE devices

I was wondering if the default behaviour for a device when POSTing should be

```
update_or_create(registration_id=registration_id, defaults={...})
```

updating the current user of a device, instead of rejecting the request due to a registration_id already being taken?

This way, if a user signs out of an app, then a different user signs in, the new user would get the push notifications?

jamaalscarlett:
Logging in and creating a new user are separate events.  If I have an application, and the user attempts to login, that should be a update (PUT, PATCH) , not a create (POST).  Handling of those cases should occur in your mobile or webapp.

chozabu:
I agree in principle - but consider this situation:

John is using an app, powered by django-push-notifications, and happily getting notifications
The device crashes, loses data, or otherwise logs john out, without removing his device information
John gives the device to Mary
Mary is now getting Johns notifications, even if she logs into the app - she no longer has Johns details but will still be receiving his notifications

Perhaps I am mis-understanding something - but this seems like a in-lib fixable situation?

jamaalscarlett:
If the device info is still intact, the device model should be updated with the correct user.  I don't disagree with the scenario, I just think that it should be fixed in the app, not the library.  Just because a problem can be fixed by a library change, does not mean that that is the correct approach.

chozabu:
but to implement that the app would either have to over-ride the current device registration routes, or create a new route that removes existing data from other users first (and use them instead of current routes)?

I mean - is there a situation where a user asks for push notifications to be sent to their device and it is better for them to be told `error: device id already registered` than to have the user with the device get their notifications?

I'll happily implement this internally if it really should not go in the lib, but it seems like a worse approach to me than just replacing an existing token if a duplicate is provided by another user?

jamaalscarlett:
My vote would be to handle it at the app level. @jleclanche any thoughts.  I will admit that many of the DRF related issues that people have found do not affect me because I implemented DRF myself before it was added to this library.  So I am open to the change if there is a desire for it.

chozabu:
Well - I'd be happy to implement at the app level, even skip implementing if I knew _why_ a device should be refused when it already exists - it feels like an anti-feature, but perhaps there is a good reason for this behaviour that I am missing?

chozabu:
So, I've had a quick glance over the relevant area of code, and my exact proposed change should be:
1. before saving, search for a device with matching registration_id
[...]

PR #345, opened 2016-12-15
Fix issue with user change on a Device (#296)

PierreF:
This is a fix for #296 . It's the [solution](https://github.com/jleclanche/django-push-notifications/issues/296#issuecomment-225462380) proposed by @chozabu with the following modification:

* Don't do repeat the create method, use self.queryset.model to access GCMDevice or APNSDevice
* Add the setting in PUSH_NOTIFICATIONS_SETTINGS
* Don't crash if "registration_id" is not in request.data :)


push_notifications/api/rest_framework.py
```diff
@@ -125,6 +128,22 @@ def has_object_permission(self, request, view, obj):
...
+			).first()
+			if instance:
+				serializer = self.get_serializer(instance, data=request.data)
+		if not serializer:
+			serializer = self.get_serializer(data=request.data)
+
+		serializer.is_valid(raise_exception=True)
+		self.perform_create(serializer)
+		headers = self.get_success_headers(serializer.data)
+		return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
```
jamaalscarlett: Looks good, but in the case of an update the http status code should be 200 or 204

PierreF:
Updated the PR to return a status code 200 in case of update (as do the default update mixing of rest framwork).

jleclanche:
Not a fan of the setting's name. Other than than, lgtm but I'll leave it up to @jamaalscarlett to decide.

PierreF:
UPDATE_ON_DUPLICATE_REG_ID ?
UPDATE_DEVICE_SAME_REG_ID  ?

If someone has a better name, I'll take it :)

jamaalscarlett:
I'm fine with UPDATE_ON_DUPLICATE_REG_ID, let's go with that one.

jleclanche:
Thanks!

PR #346, opened 2016-12-16
The special test case for explicit cerficate passing

nnseva:


nnseva:
Has been included and hardly reworked in the #348 so I am closing it to avoid unnecessary clashes

PR #347, opened 2016-12-16
Django 1 7 compatibility

nnseva:
Small changes need to be implemented to have a compatibility with Django-1.7 config (we are using it yet).

### DIFF
--- a/push_notifications/api/rest_framework.py
+++ b/push_notifications/api/rest_framework.py
@@ -121,6 +121,7 @@ class WNSDeviceSerializer(UniqueRegistrationSerializerMixin, ModelSerializer):
 		style={'input_type': 'text'},
 		required=False
 	)
+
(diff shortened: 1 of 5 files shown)


jleclanche:
I'm completely against this. Django 1.7 support [ended a year and half ago](https://www.djangoproject.com/download/). 1.8 is a LTS and even its support is ending in a few months, with the new 1.11 LTS releasing soon.

You need to upgrade to 1.8 at the very least. I intend Django 1.8 support to stay in until April, at the same time as extended support ends for upstream. It's also possible there will be nothing to actually break once we no longer test against it.

The [backwards incompatible changes](https://docs.djangoproject.com/en/1.10/releases/1.8/#backwards-incompatible-1-8) are not too severe; upgrading Django versions is never really too painful.

PR #348, opened 2016-12-16

nnseva:
This is a replacement for the pull requests #247, #259 and #260 with all-in-one commit.

old-style - one mobile application per server instance, compatibility mode
new-style static - multiple (several) mobile applications per server instance, settings file contains direct description of mobile applications credentials (key/certificate), a new mobile application requires settings patching or reloading
new-style dynamic - multiple (a lot) mobile applications per server instance, settings file contains reference to the custom model with mobile application credentials (key/certificate), a new mobile application can be created at runtime without server restart

This patch solves #130

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,8 @@ GCM, APNS or WNS devices and in the action dropdown, select "Send test message" 
 Note that sending a non-bulk test message to more than one device will just iterate over the devices and send multiple
 single messages.
 
+The application supports multiple mobile applications with separate application IDs using one server simultaneously.
+
 Dependencies
 ------------
 Django 1.8 is required. Support for older versions is available in the release 1.2.1.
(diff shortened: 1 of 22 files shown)


jleclanche:
You have whitespace diffs in `test_models.py`, can you fix that please?

jleclanche:
The `application_id` as a CharField gets a -1 from me. What's wrong with an int? If it needs to be opaque, you can make it a bigint and random, although that's out of scope for this library.

Additionally, if we're going that route, we should probably simply have an Application model, no?

nnseva:
The test_models.py has been hardly reworked across the whole file, so not only spaces have been changed there.

nnseva:
> @jleclanche > If it needs to be opaque, you can make it a bigint and random, although that's out of scope for this library.

This field totally out of the library scope and may be not only bigint, but any nature, so it has been made as a character field. For example: public key of the OAuth2 application.

Using the Application model is not necessary. In fact, you can totally ignore existence of the Application model and use some else model through settings.APNS_CERTIFICATES_MODEL, settings.GCM_API_KEYS_MODEL, settings.WNS_PACKAGE_SECURITY_IDS_MODEL etc.

Even more, the patch allows using different models for each of such data, or one for every of them. You can use static settings for certificates, model to store CGM API keys, and old-style one settings value for Windows Package security ID / secret key pair. This is not a useful case, but it shows a patch flexibility.

Such a flexibility requires some explanation in documentation and some number of popular templates to use. I am planning to create them just after the PR has been approved.

jleclanche:
Making it a charfield makes its indexing very expensive. And the field really should be indexed. Which brings me back to the original point, why not simply have an application model? The flexibility you describe is much clearer and self-documenting if you have one.

jleclanche:
@jamaalscarlett your input would be appreciated here.

README.rst
```diff
@@ -150,6 +173,74 @@ Note: gcm_send_bulk_message must be used when sending messages to topic subscrib
...
+
+Let say, the application ``applications`` contains a model ``ApplicationModel`` which contains three fields to
+reflect an application ID to the correspondent application credentials:
+
+.. code-block:: python
+
+    class ApplicationModel(models.Model):
+        application_id = models.CharField(max_length=64,unique=True)
+        gcm_api_key = models.TextField(null=True,blank=True)
+        apns_certificate = models.FileField(null=True,blank=True)
```
jaddison: As I mentioned in the previous PR, using a FileField doesn't work for horizontal scaling of servers, as the cert file would be uploaded onto a single server.

I also realize that your `ApplicationModel` here is a sample model definition - can you elaborate how the supporting code in this PR would allow the storage of the certfile contents in the DB to allow for horizontal scaling? 

In my previous #339 PR, I had some code that temporarily wrote the content out to a file, then remove it after the call succeeded.  Not efficient, but could something like this be supported? Otherwise, I'm not sure APNS in a model makes much sense. Thanks!

nnseva:
@jleclanche

It is not obviously to use the included Application model to store keys. In fact, it has been created just as exemplary and the testing feature. I would be appretiated to have any ideas from you to mark an Application model as optional and required only for tests.

The idea is that the user is free to use any type of storable data as an application_id. It is opaque for other code and just transmitted to accessor functions. It might be necessary to check additionally, whether it is possible to use integers as an application_id value.

nnseva:
@jaddison 

I've already written here that the Application model is not necessary to use. You can use any model and any fields as you wish, but you should note that the apns_certificate field should contain, or be convertible to the file path, or have an attribute`path`, or have a key`'path'` which is a real file path. So if you use some smart storage containing a file, you can implement temporary storing and caching of the file while accessing to the field value. Anyway, the real file system file presence is required just because of implementation of the SSL library used there.

jaddison:
@nnseva I understand almost everything you've said, thanks for clarifying. The one thing I can't wrap my head around is how I'd have a 'smart' storage that would have the file on disk when needed, but clean the file up afterwards (if desired).

I can totally come up with hashing the contents of the apns cert field to get a `/tmp/<hash digest>.crt` file for the ssl operation - but what I can't quite wrap my head around is how to have it automatically clean it up out of the `/tmp/` dir without race conditions (ie. the file disappearing when the ssl operation needs it).

If I understood that, then this wouldn't be an issue for me at all. :)

jaddison:
Hi everyone, just a gentle ping to see how this is coming along. Cheers!

nnseva:
> @jaddison commented 22 days ago
> ...The one thing I can't wrap my head around is how I'd have a 'smart' storage that would have the file on disk when needed, but clean the file up afterwards (if desired).

Really you should think about - when you really need to clean up the file?

1. The server needs this file every time when it connects to the push notification server, so you should not cleanup this file till you need to send notifications for this particular application. It means that you should leave the file to avoid writing it fast every time when the notification is sent.
2. In some configurations, the /tmp/ directory is cleaned up automatically when the server restarts. It means that you probably don't need to clean up the file yourself at all.
3. You probably really need to clean up the file when the record about `mobile application` is deleted.
4. See the Django-native File or Image field which are using Django file storage abstraction - as I remember, they don't delete a file from the storage when the record is deleted, why you should do it? If they do - just do it the same way. See also discussions [here](http://timonweb.com/posts/cleanup-files-and-images-on-model-delete-in-django/) and [here](http://stackoverflow.com/questions/5372934/how-do-i-get-django-admin-to-delete-files-when-i-remove-an-object-from-the-datab)

jleclanche:
@jamaalscarlett ping

push_notifications/gcm.py
```diff
@@ -47,11 +52,11 @@ def _gcm_send(data, content_type):
...
 	}
 	request = Request(SETTINGS["GCM_POST_URL"], data, headers)
-	return urlopen(request, timeout=SETTINGS["GCM_ERROR_TIMEOUT"]).read().decode("utf-8")
+	return decodestr(urlopen(request, timeout=SETTINGS["GCM_ERROR_TIMEOUT"]).read())
 
 
-def _fcm_send(data, content_type):
-	key = SETTINGS.get("FCM_API_KEY")
+def _fcm_send(data, content_type, application_id):
+	key = get_gcm_api_key(application_id)
```
Antwan86: This is not going to work. Double check the merge of your code...

nnseva: The code means that either GCM, or FCM API KEY is used for the application, not the both simultaneously.

I also really wanted to have some proper tests here. Unfortunately, I've just proposed some environment for the multiple configurations, not the functionality itself, so I am not ready to have responsibility for the functional part.

Antwan86: Yes but if you're sending a notification in bulk to a queryset with a mix of GCM/FCM applications, it's not going to work...
You made a typo it should be `get_fcm_api_key()`.

Antwan86:
(Just out of interest, I also merged the PR #352 with this PR on [this branch](https://github.com/buzzhire/django-push-notifications/tree/feature/multiple-mobile-applications-and-fcm-improvements) in case it's needed)

Antwan86:
(Updated it again using latest master + #360 in the [same branch](https://github.com/buzzhire/django-push-notifications/tree/feature/multiple-mobile-applications-and-fcm-improvements))

Antwan86:
Any update on this ?
The solution could be to use the new [APNS JWT](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW4), or simply just remove the whole Application model and possibility to have dynamic applications (which is highly unusual IMHO).

README.rst
singleton11: Agreed, you have to store APNS certificate as a text in DB

jleclanche:
@Antwan86 Support for JWT would be very nice.

push_notifications/__init__.py
```diff
@@ -1,7 +1,17 @@
...
 __version__ = "1.4.1"
 
 
 class NotificationError(Exception):
-	pass
+    pass
+
+
+if sys.version_info.major > 2:
+    def decodestr(s):
```
matthewh: @jleclanche Do you think this necessary?

jleclanche: No, it's pretty nasty.

Issue #313, opened 2016-08-12
Support for `mutable-content` key in APNS payload


evdoks:
In order to be able to customise content of notification (starting iOS 10), `mutable-content` key with the value set to 1 must be present in APNS payload (see [Apple docs](https://developer.apple.com/reference/usernotifications/unnotificationserviceextension)). Support for this key will require a minor modification to `_apns_send()` function. I could provided a PR for this feature.

jleclanche:
Go for it.

gklka:
Until that happens, we can use something like this:

```python
device.send_message("[will be overwritten]", extra={
    "aps": {
        "alert": {
            "title": "Bold text in the notification",
            "body": "Second line in the notification"
        },
        "sound": "default",
        "mutable-content": 1,
        "data": {
            "attachment-url": "https://m.popkey.co/987552/NGLb3.gif"
        }
    }
})
```

(Example for an iOS 10 rich notification)

HashimJVZ:
This is already done.

PR #350, opened 2017-01-05
Support for mutable-content flag in APNS

aaronlewism:
Support for mutable-content flag, new in iOS 10. #313


jleclanche:
Thanks.

We'll really have to rework that atrocious keyword arg list...

PR #356, opened 2017-02-18
Allow send_message to set thread-id for APNS

sww314:
Add the thread-id keyword from the [Apple Payload Guide](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html#//apple_ref/doc/uid/TP40008194-CH17-SW1). 

Includes tests and documentation in the README.


jleclanche:
👌 thank you!

PR #352, opened 2017-01-25
Consistent support for Firebase Cloud Messaging

Antwan86:
This PR adds improved support for Firebase Cloud Messaging, sending the proper messages payload. Addresses #349 #343 amongst others.

Summary
---------

Doc is updated to reflect the changes, but here is the summary :

- Any FCM device (ie with `cloud_message_type='FCM'`) will now send appropriate payload using [notification/data formated messages](https://firebase.google.com/docs/cloud-messaging/concept-options#notifications_and_data_messages).
- Valid notification parameters are automatically retrieved and send in the `notification` json object, and can be defined in two different manners :
   - Using the `extra` dictionnary : `d.send_message("msg", extra={"title":"something"})` 
   - Using directly keyword args : `d.send_message("msg", title="something")` 
- The other parameters are sent in `data` json object.
- GCM devices will keep send the former format (using `data` json object only). This can also be enforced with FCM devices passing the `use_fcm_notifications=False` options.

Cleanup / Design choices
-------------------------

1.  Readme is updated, so FCM is now mentioned everywhere and referred as the preferred method to use Google Cloud. This is however not defaulted when creating new devices, so it won't break any existing installation.
2. The `gcm.py` connector is now using exclusively `application/json` API calls, and passing a list of `registration_ids` as a target. As a consequence the unused plaintext calls using `application/x-www-form-urlencoded` are removed, as well as the unnecessary duplicate code
3. The `send_message` from `gcm.py` can now expect either a list of `registration_id`, or a single `registration_id`, and will transparently use the proper one. As a consequence this removes the needs of a duplicated `send_bulk_message` method. (For compatibility purposes it is still declared as an alias).
4. Some internal methods have been renamed (removed `gcm` or `json` mentions) to reflect their agnosticism.

Happy to discuss these choices of course, but I think they're all minor and bring cleanliness.

Issue #349, opened 2016-12-19
FCM support


danieltigse:
I just installed the last version but I can't send a message via FCM. I checked the files inside side-packages/push_notifications/gcm.py and there is nothing about GCM. Help please.

jamaalscarlett:
@danieltigse do you mean nothing about FCM?  It may not be in a released version.  Are you deploying from pypi, or building from source?

danieltigse:
Ohh I thought It was in the release version. Are you planning to release a new version? Meanwhile do you recommend me to build from source?

jamaalscarlett:
If you specifically want FCM, I would recommend waiting.  See https://github.com/jleclanche/django-push-notifications/issues/343

jleclanche:
Merging into #343

Issue #343, opened 2016-12-02
Wrong implementation of FCM


WarmongeR1:
Hello,

I think your implementation of FCM (#302) is wrong.

2 reasons:

First:
In docs we have requests with Content-Type: application/json
Your implementation - application/x-www-form-urlencoded;charset=UTF-8
Docs - https://firebase.google.com/docs/cloud-messaging/send-message

Second:
A body is wrong - FCM expects JSON with keys: to, data, notification. But you send dict with key registration_id

Docs - https://firebase.google.com/docs/cloud-messaging/concept-options#notifications_and_data_messages

As result, I wrote trash-implementation in commit https://github.com/WarmongeR1/django-push-notifications/commit/7fdad2a2079db1049c7c807b287d62672bf5e611
I didn't fix tests, I only demonstrate work version of code

And current interface `send_message` is unsuitable for FCM.
Because if we send Push-notification with FCM we should write

`python
device.send_message(None, extra={'notification': {'body': 'lfdlfds'}})
`

Sorry for my bad english.

jamaalscarlett:
@WarmongeR1 thanks for this.  I think that while GCM has been deprecated, Google has designed the FCM to accept GCM style messages.  Here is there troubleshooting curl command from the FCM page: [FCM server](https://firebase.google.com/docs/cloud-messaging/server)

Checking the validity of a server key
If you receive authentication errors when sending messages, check the validity of your Server key. For example, on Android, run the following command:

```
api_key=YOUR_SERVER_KEY
```
```
curl --header "Authorization: key=$api_key" \
       --header Content-Type:"application/json" \
       https://fcm.googleapis.com/fcm/send \
       -d "{\"registration_ids\":[\"ABC\"]}"
```
As you see, it uses the old GCM style request structure.
So, while #336 does work, I would agree that the implementation is not correct.
 Please note that plaint text requests are supported.  Again from the FCM docs:

Authentication

To send a message, the app server issues a POST request. For example:

https://fcm.googleapis.com/fcm/send
A message request consists of two parts: the HTTP header and the HTTP body.

The HTTP header must contain the following headers:

Authorization: key=YOUR_SERVER_KEY 
Make sure this is the server key, whose value is available in the Cloud Messaging tab of the Firebase console Settings pane. Android, iOS, and browser keys are rejected by FCM.
**Content-Type: application/json for JSON; application/x-www-form-urlencoded;charset=UTF-8 for plain text.** 
[...]

PR #360, opened 2017-02-21
Consistent support for Firebase Cloud Messaging

Antwan86:
_[Replace #352 with a better commit history]_

This PR adds improved support for Firebase Cloud Messaging, sending the proper messages payload. Addresses #349 #343 amongst others.

Summary
---------

Doc is updated to reflect the changes, but here is the summary :

- Any FCM device (ie with `cloud_message_type='FCM'`) will now send appropriate payload using [notification/data formated messages](https://firebase.google.com/docs/cloud-messaging/concept-options#notifications_and_data_messages).
- Valid notification parameters are automatically retrieved and send in the `notification` json object, and can be defined in two different manners :
   - Using the `extra` dictionnary : `d.send_message("msg", extra={"title":"something"})` 
   - Using directly keyword args : `d.send_message("msg", title="something")` 
- The other parameters are sent in `data` json object.
- GCM devices will keep send the former format (using `data` json object only). This can also be enforced with FCM devices passing the `use_fcm_notifications=False` options.

Cleanup / Design choices
-------------------------

1.  Readme is updated, so FCM is now mentioned everywhere and referred as the preferred method to use Google Cloud. This is however not defaulted when creating new devices, so it won't break any existing installation.
2. The `gcm.py` connector is now using exclusively `application/json` API calls, and passing a list of `registration_ids` as a target. As a consequence the unused plaintext calls using `application/x-www-form-urlencoded` are removed, as well as the unnecessary duplicate code
3. The `send_message` from `gcm.py` can now expect either a list of `registration_id`, or a single `registration_id`, and will transparently use the proper one. As a consequence this removes the needs of a duplicated `send_bulk_message` method. (For compatibility purposes it is still declared as an alias).
4. Some internal methods have been renamed (removed `gcm` or `json` mentions) to reflect their agnosticism.

Happy to discuss these choices of course, but I think they're all minor and bring cleanliness.


jleclanche:
Looks great! @jamaalscarlett can you review it as well?

jamaalscarlett:
:+1:

jamaalscarlett:
merged into master

PR #361, opened 2017-02-21
Replace distutils with setuptools

matthewh:
`distutils.core.setup` does not support `python setup.py develop`. Replace with the standard `setuptools.setup` variant makes developing and integrating multiple python modules easier. I also thought distutils was deprecated in favor of setuptools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 import os.path
 import push_notifications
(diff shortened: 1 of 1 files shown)


jleclanche:
Thanks, looks good, but can you fix your author email please?

matthewh:
@jleclanche I've updated my author email.

PR #364, opened 2017-02-25
Because NotRegistered and InvalidRegistration do not throw GCMError

jamaalscarlett:
check for them in django admin to correctly display error message.

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -44,6 +44,15 @@ class DeviceAdmin(admin.ModelAdmin):
 			if bulk:
 				break
 
+		# Because NotRegistered and InvalidRegistration do not throw GCMError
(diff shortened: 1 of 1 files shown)


push_notifications/admin.py
```diff
@@ -44,6 +44,16 @@ def send_messages(self, request, queryset, bulk=False):
 			if bulk:
 				break
 
+		# Because NotRegistered and InvalidRegistration do not throw GCMError
+		# catch them here to display error msg.
+		if not bulk:
+			ret_copy = list(ret)
```
jamaalscarlett: I make the copy because I am removing elements from the list in the for loop.

jleclanche: `for r in list(ret)` will still make a copy.

PR #367, opened 2017-03-01
Fix bug in UniqueRegistrationSerializerMixin.

jamaalscarlett:
The the request is a PATCH, attrs may not contain id or reg_id.
So use the values from self.instance instead.


jleclanche:
gtm

PR #370, opened 2017-03-04
Enable multiple application support

matthewh:
Provides backwards compatibility through conf adapters that determine which settings are supported and where the data can be loaded from. This implementation adds three:

  * LegacyConfig - supports the existing settings format
  * AppConfig - supports a dictionary of application_ids and their settings
  * AppModelConfig - supports storing the settings in the database for multiple apps

@jleclanche Consider this a really rough draft. I haven't had much time to test it though the tests pass.  I hope this gives you an idea of what I intend and should replace the other mutli application PR. This one has stubs for the multi-app config and multi-app in a model config. I'd like to get this approved and into the merged with the legacy config adapter just to reduce the number of rebases.

### DIFF
--- a/push_notifications/apns.py
+++ b/push_notifications/apns.py
@@ -13,7 +13,7 @@ from apns2 import payload as apns2_payload
 from . import models
 from . import NotificationError
 from .apns_errors import reason_for_exception_class
-from .settings import PUSH_NOTIFICATIONS_SETTINGS as SETTINGS
+from .conf import get_manager
 
 
 class APNSError(NotificationError):
@@ -30,20 +30,21 @@ class APNSServerError(APNSError):
 		self.status = status
 
 
-def _apns_create_socket(certfile=None):
-	certfile = certfile or SETTINGS.get("APNS_CERTIFICATE")
+def _apns_create_socket(certfile=None, application_id=None):
+	certfile = certfile or get_manager().get_apns_certificate(application_id)
 	client = apns2_client.APNsClient(
 		certfile,
-		use_sandbox=SETTINGS.get("APNS_USE_SANDBOX"),
-		use_alternative_port=SETTINGS.get("APNS_USE_ALTERNATIVE_PORT"))
+		use_sandbox=get_manager().get_apns_use_sandbox(application_id),
+		use_alternative_port=get_manager().get_apns_use_alternative_port(application_id)
+	)
 	client.connect()
 	return client
 
 
 def _apns_prepare(
-	token, alert, badge=None, sound=None, category=None, content_available=False,
-	action_loc_key=None, loc_key=None, loc_args=[], extra={}, mutable_content=False,
-	thread_id=None, url_args=None):
+	token, alert, application_id=None, badge=None, sound=None, category=None,
+	content_available=False, action_loc_key=None, loc_key=None, loc_args=[],
+	extra={}, mutable_content=False, thread_id=None, url_args=None):
 		if action_loc_key or loc_key or loc_args:
 			apns2_alert = apns2_payload.PayloadAlert(
 				body=alert if alert else {}, body_localized_key=loc_key,
@@ -59,8 +60,10 @@ def _apns_prepare(
 			url_args, custom=extra, thread_id=thread_id)
 
 
-def _apns_send(registration_id, alert, batch=False, **kwargs):
-	client = _apns_create_socket(kwargs.pop("certfile", None))
+def _apns_send(
+	registration_id, alert, batch=False, application_id=None, certfile=None, **kwargs
+):
+	client = _apns_create_socket(certfile=certfile, application_id=application_id)
 
 	notification_kwargs = {}
 
@@ -80,14 +83,19 @@ def _apns_send(registration_id, alert, batch=False, **kwargs):
 		data = [apns2_client.Notification(
 			token=rid, payload=_apns_prepare(rid, alert, **kwargs)) for rid in registration_id]
 		return client.send_notification_batch(
-			data, SETTINGS.get("APNS_TOPIC"), **notification_kwargs)
+			data, get_manager().get_apns_topic(application_id=application_id),
+			**notification_kwargs
+		)
 
 	data = _apns_prepare(registration_id, alert, **kwargs)
 	client.send_notification(
-		registration_id, data, SETTINGS.get("APNS_TOPIC"), **notification_kwargs)
+		registration_id, data,
+		get_manager().get_apns_topic(application_id=application_id),
+		**notification_kwargs
+	)
 
 
-def apns_send_message(registration_id, alert, **kwargs):
+def apns_send_message(registration_id, alert, application_id=None, certfile=None, **kwargs):
 	"""
 	Sends an APNS notification to a single registration_id.
 	This will send the notification as form data.
@@ -100,7 +108,10 @@ def apns_send_message(registration_id, alert, **kwargs):
 	"""
 
 	try:
-		_apns_send(registration_id, alert, **kwargs)
+		_apns_send(
+			registration_id, alert, application_id=application_id,
+			certfile=certfile, **kwargs
+		)
 	except apns2_errors.APNsException as apns2_exception:
 		if isinstance(apns2_exception, apns2_errors.Unregistered):
 			device = models.APNSDevice.objects.get(registration_id=registration_id)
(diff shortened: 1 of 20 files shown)


tests/test_gcm_push_payload.py
```diff
@@ -25,21 +37,21 @@ def test_fcm_push_payload_params(self):
 				b'{"data":{"other":"misc"},"delay_while_idle":true,'
 				b'"notification":{"body":"Hello world","title":"Push notification"},'
 				b'"registration_ids":["abc"],"time_to_live":3600}',
-				"application/json")
+				"application/json", application_id=None)
```
jleclanche: Isn't it None by default?

matthewh: 1. send_bulk_message is called without setting the application_id. 
1. send_bulk_message is really send_message.
1. _cm_send_request is called by send_message with an explicit application_id=application_id, where application_id=None. So, yes it's None by default on all of these methods but, because they call each other, it is explicitly set to None by the callee.

matthewh:
@jleclanche I'm getting setup with WNS and FCM accounts to test these changes against real hosts. Do you have any additional feedback on what's here?

jleclanche:
Not so far, looks great overall. @jamaalscarlett @kit-cat @Proper-Job @Antwan86 y'all wanna give some feedback?

push_notifications/admin.py
```diff
@@ -85,11 +85,10 @@ def prune_devices(self, request, queryset):
 		# if the user doesn't select all the devices for pruning, we
 		# could very easily leave an expired device as active.  Maybe
 		#  this is just a bad API.
-		expired = get_expired_tokens()
-		devices = queryset.filter(registration_id__in=expired)
-		for d in devices:
-			d.active = False
-			d.save()
+		app_ids = set(queryset.values_list('application_id', flat=True).distinct())
```
kit-cat: I don't think the `set()` call is needed.

push_notifications/admin.py
```diff
@@ -85,11 +85,10 @@ def prune_devices(self, request, queryset):
...
+		for app_id in app_ids:
+			expired = get_expired_tokens(app_id)
+			queryset.filter(registration_id__in=expired).update(active=False)
```
kit-cat: I would also add app_id in the filter: `filter(registration_id__in=expired, application_id=app_id)`

push_notifications/conf/app.py
```diff
@@ -0,0 +1,71 @@
...
+PUSH_NOTIFICATIONS_SETTINGS = {
+	# defines where to load application configuration from,
+	# default will be the old file format from settings.py (LegacySettingsAdapter).
+	"CONFIG": "push_notifications.conf.AppConfig"
+
+	# new array of all applications in the push_notifications
+	"APPLICATIONS": {
+		# required, how to address messages to this client from code
+		"my_fcm_app": {
+			# required - indicates what fields are needed to get api key or details: FCM, GCM, IOS, WNS
```
kit-cat: I would use APNS, not IOS, for consistency

push_notifications/models.py
```diff
@@ -49,16 +54,18 @@ def send_message(self, message, **kwargs):
 			if message is not None:
 				data["message"] = message
 
+			app_ids = set(self.filter(active=True).values_list('application_id', flat=True).distinct())
```
kit-cat: Same here - I think the `set` call is unneeded

push_notifications/models.py
```diff
@@ -98,11 +105,20 @@ def get_queryset(self):
...
 
 class APNSDeviceQuerySet(models.query.QuerySet):
-	def send_message(self, message, **kwargs):
+	def send_message(self, message, certfile=None, **kwargs):
 		if self:
 			from .apns import apns_send_bulk_message
-			reg_ids = list(self.filter(active=True).values_list("registration_id", flat=True))
-			return apns_send_bulk_message(registration_ids=reg_ids, alert=message, **kwargs)
+
+			app_ids = set(self.filter(active=True).values_list('application_id', flat=True).distinct())
```
kit-cat: Same

push_notifications/models.py
```diff
@@ -135,8 +156,16 @@ def send_message(self, message, **kwargs):
 		if self:
 			from .wns import wns_send_bulk_message
 
-			reg_ids = list(self.filter(active=True).values_list("registration_id", flat=True))
-			return wns_send_bulk_message(uri_list=reg_ids, message=message, **kwargs)
+			app_ids = set(self.filter(active=True).values_list('application_id', flat=True).distinct())
```
kit-cat: same

push_notifications/apns.py
```diff
@@ -64,8 +65,8 @@ def _check_certificate(ss):
 		raise ImproperlyConfigured("The APNS certificate doesn't contain a private key")
 
 
-def _apns_create_socket(address_tuple, certfile=None):
-	certfile = certfile or SETTINGS.get("APNS_CERTIFICATE")
+def _apns_create_socket(address_tuple, application_id=None, certfile=None):
+	certfile = certfile or manager.get_apns_certificate(application_id)
 	if not certfile:
 		raise ImproperlyConfigured(
 			'You need to set PUSH_NOTIFICATIONS_SETTINGS["APNS_CERTIFICATE"] to send messages through APNS.'
```
kit-cat: `manager.get_apns_certificate` will already raise - we can remove those lines

push_notifications/admin.py
matthewh: That's definitely a mistake. Good eyes.

push_notifications/conf/app.py
matthewh: Agreed. I'll switch the platform name from IOS to APNS.

matthewh:
@kit-cat I've removed all the redundant set()s.

matthewh:
My next step is to write tests for the AppConfig. I can commit as I go or hold off till its further along if you prefer. Thanks for the feedback so far.

kit-cat:
Thank you @matthewh ! I don't really have a preference - depends on what @jleclanche thinks.

jleclanche:
I don't mind either way; you can keep working on this PR. Travis needs to be fixed btw.

matthewh:
@jleclanche Rebased and flaked. Still testing.

andreyshipilov:
Looks like tests are passing now. We just got to this point in our app that we need multi-app support, so would be super sweet to have it in the main repo as soon as possible :)

jleclanche:
Went over the PR, looks good. @matthewh I also invited you as a collaborator. You can land this PR once you're comfortable with it as well.

Please do keep PRing your (non-trivial) changes though :)

matthewh:
Thank you @jleclanche. I have iOS and Android fairly well tested. I got hung up on setting up an WNS app and need to revisit it. I have another set of updates to land tomorrow for review.

matthewh:
@jleclanche I consider this the final revision save any bugs I find testing WNS or issues you (or others) raise.

push_notifications/conf/app.py
```diff
@@ -0,0 +1,357 @@
+from django.core.exceptions import ImproperlyConfigured
+from django.utils.six import string_types
+from .base import BaseConfig, check_apns_certificate
+from ..settings import PUSH_NOTIFICATIONS_SETTINGS as SETTINGS
+
+"""
```
jleclanche: Could you move this massive docstring somewhere it's more likely to be seen by a user?

We don't have doc pages right now, so either:

- The wiki
- The README.rst (though it'd have to be more terse)
- An example settings file you ship alongside the rest of the code.

matthewh: How about all three? The README should at least contain a shoutout and link to the wiki. Where in the source would you imagine shipping this example config? Would dropping it at tests/settings/app_config.py make sense?

jleclanche: I sometimes see a pattern of shipping an example project, [like in allauth](https://github.com/pennersr/django-allauth). Don't sweat it though.

matthewh: An example settings file in the source seems problematic since the user would need to provide api keys or certificates. I moved that massive docstring to the [wiki](https://github.com/jleclanche/django-push-notifications/wiki/Multiple-Application-Support) and expanded it.

andreyshipilov:
Arrr, conflicts :(

matthewh:
yeah yeah :) I have fixed the conflicts but there are some other setting changes that now need to be made. I'm working on it.

matthewh:
This is finally ready to go. Please review one more time and let's mainline this.

andreyshipilov:
Oh, lord we so need this over here :)

jleclanche:
This looks good to land! It doesn't have to be in this PR specifically but please write a little upgrade path somewhere for those who want to use the new system.

jamaalscarlett:
https://github.com/jleclanche/django-push-notifications/pull/348 should probably be closed now.  Great work @matthewh

PR #371, opened 2017-03-04
Fix fcm/gcm send_message method.

### LINKED ISSUE #369: send_message can't send push-notification to the topic.
It seems to me, that current version of `send_message` function, doesn't have the possibility to send message to the topic. The `registration_ids` parameter should be None, when you want to send something on topic, but then any `_cm_send_request` (etc) is not running.

jamaalscarlett:
The refactored send_message method required a registration_id thus
breaking the ability to send messages to topics.
Fixes #369


jleclanche:
fix flake8 and gtm

Issue #216, opened 2015-09-11
Disable duplicates of registration_id


saritasa:
Right now if I can add unlimited amount of duplicates and then when I'm using bulk send_message I'm getting duplicated push notifications equal to quantity of dups inside DB. Would be nice to prevent such thing and don't add duplicate value to DB

jamaalscarlett:
What value is duplicate?  The registration_id?

saritasa:
correct. It's possible to commit multiple duplicates and when do send_message bulk - it sends a lot of dups

jamaalscarlett:
But where are the duplicates coming from?  You don't create the registration_id yourself, those come from google.

saritasa:
Example for APNS:

```
POST /api/v1/devices/apns/ HTTP/1.1
Host: domain.com
Content-Type: application/json
Authorization: token 2192967b2fc60623597790783c9590a1964c6e62
Cache-Control: no-cache
{
    "name": "iOS 6.1",
    "registration_id": "4d2d8feaf9a2d780c6c2d18a83c94483462ebe74aafab1f989711e6292b96c9b"
}
```

**response:**

```
{
  "registration_id": [
    "This field must be unique."
  ]
}
```

**example of GCM**

```
POST /api/v1/devices/gcm/ HTTP/1.1
Host: domain.com
Content-Type: application/json
Authorization: token 2192967b2fc60623597790783c9590a1964c6e62
Cache-Control: no-cache

{
    "name": "Android 5.0",
    "registration_id": "APA91bF2FA873EZ8vv-JifiSKPP8wxBCxOfPhLaHkH2d9mYC-CEY_0I0ZRxRPrGqCdCgMybKUPtGKDqNN8ZgTdtHHp7IGdY5DNqzLUK2ylLpffclz-K4MS80XG16pV-DCuzi4EdRltMp"
}
```

**response:**
```
{
  "name": "Android 5.0",
  "registration_id": "APA91bF2FA873EZ8vv-JifiSKPP8wxBCxOfPhLaHkH2d9mYC-CEY_0I0ZRxRPrGqCdCgMybKUPtGKDqNN8ZgTdtHHp7IGdY5DNqzLUK2ylLpffclz-K4MS80XG16pV-DCuzi4EdRltMp",
  "device_id": "",
  "active": true,
  "date_created": "2015-09-18T21:58:35.788556Z"
}
```
**Another call with the same ID:**

```
{
  "name": "Android 5.0",
[...]

Issue #366, opened 2017-02-27
Application Support 1.5 Proposal


matthewh:
This is an intermediate step before implementing full in database support for multiple applications. 

* Move drf and tastypie integrations into push_notifications.contrib. Add deprecated warnings at their current location for backwards compatibility.
* Add support for configuration adapters.
    * Add a LegacySettingsAdapter that supports the existing format. (this will be default)
    * Add a AppIDSettingsAdapter that address applications by application_id for all platforms. (this will be default)
* Add support for application_id based on the current PR but skip any "other" features at this time.

This proposal suggests an alternate settings format as well as formally supporting additional contributed features via a push_notifications.contrib module. The drf and tastypie integrations would be moved here (with links maintained in their old locations for backwards compatibility). 

Additionally, the registry module that provides mobile clients to register tokens and other backend services to send push notifications through an API would be under push_notifications.contrib. (more to come on registry in another proposal).

```
PUSH_NOTIFICATIONS_SETTINGS = {
    # defines where to load application configuration from,
    # default will be the old file format from settings.py (LegacySettingsAdapter).
    "CONFIG_ADAPTER": "push_notifications.adapters.AppIDSettingsAdapter"

    # new array of all applications in the push_notifications
    "APPLICATIONS": [
        {
            # required - indicates what fields are needed to get api key or details: FCM, GCM, IOS, WNS
            "PLATFORM": "FCM",

            # required, how to address messages to this client from code
            "APPLICATION_ID": "my_fcm_app" 

            # optional, will be "default", allows for addressing devices across platform boundaries.
            # For instance, I have an android and ios version of my application
            # and I'd like to notify all users of a special announcement.
            "APPLICATION_GROUP": "my_application",

            # optional. if set, integrates with contrib.registry to allow mobile
            # applications to add keys (in situations where user authentication is not desired)
            "APPLICATION_SECRET": "d0910e22bd008b8ef603f5432d1e7020568ca0e9"

            # fcm setting
            "API_KEY": "[your api key]",
        },
        {
            "PLATFORM": "IOS",  # required - indicates what fields are needed to get api key or details: FCM, GCM, IOS, WNS

            # required, how to address messages to this client from code
            "APPLICATION_ID": "my_ios_app" 

            # optional, will be "default", allows for addressing devices across platform boundaries.
            # For instance, I have an android and ios version of my application
            # and I'd like to notify all users of a special announcement.
            "APPLICATION_GROUP": "my_application",

[...]

Issue #170, opened 2015-06-25
New APNS API announced at WWDC 2015


chrisballinger:
Just a FYI there will be a big change in APNS soon: https://developer.apple.com/videos/wwdc/2015/?id=720

The new API is way cleaner, backwards compatible, and fixes a lot of the flaws from the original design (especially error handling).

jleclanche:
Sounds great.

Alp-Phone:
The relevant talk starts at 16:45 minutes.
Thanks for the link, it looks like it'll fix all our current problems :)

imkevinxu:
Is there any work being done in prep for the new APNS API?

jleclanche:
Not from me, but I'm welcome to PRs.

Alp-Phone:
I haven't started working on this either.  Is there a release date?
I do want to be involved so if somebody is planning to work on this feature, give me a ping.

jleclanche:
Subscribe to this issue, this will be the tracker for it.

Alp-Phone:
Done

koutoftimer:
@Proper-Job there are a lot of work to do. 
First of all, I would like to move all code related to sending push notifications to external library like [apns-client](https://bitbucket.org/sardarnl/apns-client) and [gcm-cliend](https://bitbucket.org/sardarnl/gcm-client/). I'm not saying they are perfect solutions, I'll do some research on this topic.

@jleclanche nothing personal, but I think it is better to develop libraries that can handle sending push notifications for everybody who use python, and then use them for building library for django notifications.

Proper-Job:
@koutoftimer If you need an extra pair of eyes, I'll be happy to review changes.

jleclanche:
> I think it is better to develop libraries that can handle sending push notifications for everybody who use python, and then use them for building library for django notifications.

I'm in favour of that approach, but when I created this library there was nothing like that then. I'll be happy to discuss this further later.

jleclanche:
PyAPNS2 takes care of this.

PR #375, opened 2017-03-20
Allow device uniqueness to be configured via settings

matthewh:
This will allow application developers to determine the best approach for device
 uniqueness in their application.

* For backwards compatibility we keep the registration_id on APNSDevice unique.
* The check can be overriden with a callable or tuple via UNIQUE_DEVICE_CHECK.

Resolves: #216


matthewh:
I thought the cleanest approach to supporting customizable device unique requirements was to overload the Device models [_get_unique_checks](https://github.com/django/django/blob/8346680e1ca4a8ddc8190baf3f5f944f6418d5cf/django/db/models/base.py#L944) to inject a custom unique_together constraint.

jleclanche:
I think it's a very bad idea to implement a setting for this. Migrations and settings never play well together.

I think the real approach is to allow subclassing of the *Device classes. But this needs more discussion before we let it go in.

In any case, at this point I don't want to take in any other rework until 1.5 is releasable. I really want 1.5 to get released and then we can talk about landing stuff for 2.0.

matthewh:
Django unique_together constraints are applied at the application, not database level. This solution neatly sidesteps any migration issues by avoiding Meta.unique_together (which would generate a migration) and applies the constraint only at runtime. Device subclassing is definitely workable too. 

Is the multi-application PR on the table for 1.5?

jleclanche:
> Django unique_together constraints are applied at the application, not database level

Is that so? That doesn't sound right at all.

https://docs.djangoproject.com/en/1.10/ref/models/options/#unique-together

> This is a tuple of tuples that must be unique when considered together. It’s used in the Django admin and is enforced at the database level (i.e., the appropriate UNIQUE statements are included in the CREATE TABLE statement).

jleclanche:
> Is the multi-application PR on the table for 1.5? Do you have a written roadmap for 1.5? I have this PR, the multiple-app PR, and a registry PR waiting in the wings.

For 1.5 I want @kit-cat's PR to land to move us to pyapns2, which should get us #170 and #313 (or close enough to it that we can squash them). I also want #366 to land. I think that's enough for it. Everything else is in the [2.0 milestone](https://github.com/jleclanche/django-push-notifications/issues?q=is%3Aopen+is%3Aissue+milestone%3A2.0).

I'm also not opposed to releasing a 1.6 in between 1.5 and 2.0 if we see that the backwards-incompatible changes are taking a long time to land.

matthewh:
True, but Django does not rely on the database raising a constraint error. Django checks these constraints manually at the application level:

https://github.com/django/django/blob/8346680e1ca4a8ddc8190baf3f5f944f6418d5cf/django/db/models/base.py#L995

matthewh:
Closing, does not meet package maintainer's requirements.

PR #386, opened 2017-04-17
GCM/FCM device uuid

matthewh:
Replace HexIntegerField with UUIDField. Allows backwards compatibility through the REST API by left zero-padding values.


matthewh:
@jleclanche I'm on the fence about making the GCM device_id a UUID. I think an CharField that can accommodate a UUID would be more appropriate for backwards compatibility.

The [guidelines](https://developer.android.com/training/articles/user-data-ids.html) indicate a UUID is appropriate but I don't have enough experience with the alternatives. If the device_id is to be useful when sending messages with a query then we shouldn't mess with the sent value too much or the developer will need to remember to translate the device_id when querying. Seems error prone.

jleclanche:
That page you linked talks about creating a random uuid.

I think it's fine, tbh, but the main things that need to be tested are how the DRF API reacts to it, and how the migration holds up on an existing db.

matthewh:
@jleclanche If we adopt UUIDs (and the provided UUIDField for storage and serialization) there is no way to support old data, even with a migration. Existing GCM apps would have to update their logic to send UUIDs because the UUIDFields expect full uuid values. To maintain backwards compatibility, and for the migrations data to work, we must provide a custom serialization uuid field that can accept the old values and left pad them before they are saved to the database.

jleclanche:
A data migration is all that's needed. Left-padding isn't necessary, UUIDs are just 128-bit ints. What's the issue exactly?

matthewh:
Django's UUIDField only doesn't accept some value types we could expect from Android apps that aren't using UUIDs.

Existing installations are likely to try values like ``0xdeadbeaf`` and fail without a custom UUID Serializer in DRF. Are you ok with that?

```
		# device_id is optional - success
		device = GCMDevice.objects.create(
			registration_id="foobar",
			device_id=None
		)

		# short values not supported - raises ValueError
		with self.assertRaises(ValueError) as raised:
			device = GCMDevice.objects.create(
				registration_id="foobar",
				device_id=123456
			)

		# short values not supported - raises ValueError
		with self.assertRaises(ValueError) as raised:
			device = GCMDevice.objects.create(
				registration_id="foobar",
				device_id="0xdeadbeaf"
			)
```

jleclanche:
Are you talking about specifically DRF? If so, yes, I'm fine with us using a custom serializer wrapping UUIDField.

yauheni-holik:
Is this going to be ever merged? It seems like a valid solution

ddemid:
Is this going to be ever merged? it's seems like a valid solution

acenario:
Bump!

auvipy (changes requested):
this needs rebase

PR #400, opened 2017-05-05
Fix 'APNS_TOPIC' docs

kit-cat:
Passing `topic=...` to PyAPNS2 is mandatory (see https://github.com/Pr0Ger/PyAPNs2/issues/29), the topic isn't auto-guessed from the certificate.
An emtpy APNS_TOPIC setting leads to TopicDisallowed / MissingTopic errors.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,7 +72,7 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
 **APNS settings**
 
 - ``APNS_CERTIFICATE``: Absolute path to your APNS certificate file. Certificates with passphrases are not supported.
-- ``APNS_TOPIC``: The topic of the remote notification, which is typically the bundle ID for your app. If you omit this header and your APNs certificate does not specify multiple topics, the APNs server uses the certificate’s Subject as the default topic.
+- ``APNS_TOPIC``: The topic of the remote notification, which is typically the bundle ID for your app.
 - ``APNS_USE_ALTERNATIVE_PORT``: Use port 2197 for APNS, instead of default port 443.
 - ``APNS_USE_SANDBOX``: Use 'api.development.push.apple.com', instead of default host 'api.push.apple.com'.
 


matthewh:
@kit-cat Apple's documentation for [apns-topic](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html) states that it will use the certificate's Subject if apns-topic is not provided. Is the issue here that pyapns2 is always explicitly setting the header even if the value is empty? Would it be appropriate to open a PR on pyapns2 to address this?

kit-cat:
@matthewh : Yes, exactly - see https://github.com/Pr0Ger/PyAPNs2/blob/ae835a195f5ed06004604503ecaf11d20cce6128/apns2/client.py#L63, for instance
PyAPNS2 did have optional topics at first, but they switched to required topics after https://github.com/Pr0Ger/PyAPNs2/issues/29 - I'm not sure they would be open to reverting to optional topics, but we can try making a PR on pyapns2 to address this

matthewh:
I find pyapns2 presumption that the topic should be required misguided. The method happily takes **None** and an empty string as input but those will never have the desired effect. Either of those could be construed by a developer as omitting the topic. This is what I believed.

If the topic was optional as Apple allows, would developers who upgraded to Django Push Notifications 1.5 have been able to continue working without changing their settings. We have strived to maintain backwards compatibility without forcing developers to tweak settings/etc. In this regard I feel we have let our developer base down. If an optional topic would not improve the situation for developers upgrading to 1.5 with the older certificates then it doesn't matter.

kit-cat:
> I find pyapns2 presumption that the topic should be required misguided.

Yes, I agree 100%.

> If the topic was optional as Apple allows, would developers who upgraded to Django Push Notifications 1.5 have been able to continue working without changing their settings. We have strived to maintain backwards compatibility without forcing developers to tweak settings/etc. In this regard I feel we have let our developer base down. If an optional topic would not improve the situation for developers upgrading to 1.5 with the older certificates then it doesn't matter.

"Older certificates" will still work with a 1.5 upgrade without specifying the topic if PyAPNS2 lets us have optional topics again. Something like:

```
headers = {}
if topic:
    headers['apns-topic'] = topic
```

instead of https://github.com/Pr0Ger/PyAPNs2/blob/master/apns2/client.py#L63 would work and we would be able to send notifications without requiring our users to specify APNS_TOPIC again.

kit-cat:
In the meantime, we still need to fix the docs - we can revert this when APNS_TOPIC is optional again

Pr0Ger:
Hello,

Yes, `apns-topic` header is optional for new APNs API, but omitting this header has some limitations:
1) you must specify it if you use JWT tokens instead of certificates
2) you must specify it if your certificate has support for multiple topics.

All production certificates generated since the introduction of new API (which happens at the end of 2015) have three support for three topics (`bundle_id`, `bundle_id.voip`, `bundle_id.complication`). So right now there are no valid production certificates which can be used without sending `apns-topic` header and only sandbox-only certificates can be used without sending this header.

I'm not against changing this parameter back to optional, but trying to send notification without topic is most likely an error.

Pr0Ger:
Oh, I forget about push notifications for Safari, even if nobody is using it, it's possible to send these pushes without specifying topic in the production environment. I'll fix it soon

kit-cat:
Thank you for your feedback, @Pr0Ger! Happy to hear specifying topics will be optional again :slightly_smiling_face:

PR #403, opened 2017-05-17
Remove push_notifications.utils

blueyed:
It only contains `get_expired_tokens`, which is not importable anymore,
since apns_fetch_inactive_ids has been removed in 3998996.

Ref: https://github.com/jleclanche/django-push-notifications/issues/389


jleclanche:
Thank you!

PR #406, opened 2017-05-30
Fixes error when sending django admin test gcm message.

jamaalscarlett:
Fixes display issue when sending test bulk message > max_recipients
Adds original_registration_id to results dict on GCMError

Addresses:
https://github.com/jleclanche/django-push-notifications/issues/354
&
https://github.com/jleclanche/django-push-notifications/issues/391

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -47,21 +47,31 @@ class DeviceAdmin(admin.ModelAdmin):
 		# Because NotRegistered and InvalidRegistration do not throw GCMError
 		# catch them here to display error msg.
 		if not bulk:
-			for r in list(ret):
-				if r in ("Error=NotRegistered", "Error=InvalidRegistration"):
-					errors.append(r)
-					ret.remove(r)
+			for r in ret:
+				if "error" in r["results"][0]:
+					errors.append(r["results"][0]["error"])
 		else:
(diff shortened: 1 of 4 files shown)


push_notifications/admin.py
```diff
@@ -47,21 +47,31 @@ def send_messages(self, request, queryset, bulk=False):
...
 				level=messages.ERROR
 			)
 		if ret:
-			if not bulk:
-				ret = ", ".join(ret)
-			elif ret[0][0]["success"] == 0:
+			if bulk:
+				try:
+					success = ret[0][0]["success"]
+				except TypeError:
```
jamaalscarlett: When the queryset exceeds the max_recipients value, the send_message method returns a list of dicts, one per chunk.

jamaalscarlett: Ideally, the responses would be combined, but currently they are not, and I did not want to break backwards compatibility.

jleclanche: Can you add a comment explaining what's going on? What is the structure of the two different possibilities we should expect?

A test for both would be welcome.

jamaalscarlett: Will do, there are a bunch of failing tests I still need to fix.

jamaalscarlett:
@jleclanche I think this is ready.

PR #417, opened 2017-07-23
admin fix, unable to modify device entry, getting error about Hex field

theromis:


jamaalscarlett:
@theromis this fix is not valid.  The device_id field is not read-only and should not be read only in the django admin.

theromis:
@jamaalscarlett In this case please tell me how I can edit GCM devices in admin, because when I'm saving device entry I'm getting error.

jamaalscarlett:
@theromis  Which version are you using?  And what device_id value are you attempting to save?  A stacktrace would also be helpful.

theromis:
my version is 1.5.0
capture

jamaalscarlett:
@theromis what is the value you are trying to save?

theromis:
`15294002812468627586`

jamaalscarlett:
Unfortunately this is a known issue with the HexIntegerField.  There is an existing PR that should fix this: https://github.com/jleclanche/django-push-notifications/pull/386

theromis:
Ah, in this case great, waiting for this PR approval.

PR #426, opened 2017-09-27
Support for JWT Auth

shigmas:
I've had this sitting locally for a while, but I'd like to push this out, so best to use the upstream repo.

This keeps the same API, allowing users to continue to use either certificates or JWT auth. The plus for this is that you can use one Auth Token for all applications, rather than a certificate for each. This simplifies configuration immensely.


shigmas:
I'll squash my commits for the flake8 errors after comments, but it looks like we'll have to exclude Django 2.x to fix the other error.

push_notifications/models.py
```diff
@@ -109,12 +109,17 @@ def send_message(self, message, **kwargs):
 
 
 class APNSDeviceManager(models.Manager):
+	# Here's where we could keep the credentials for a JWT so we don't have
+	# to recreated it for every request
+	#
+	# self.credentials = ...
+	# This would require some API to create some credentials maybe...
```
jleclanche: Please remove this comment; that's most likely not going to happen.

push_notifications/conf/app.py
```diff
@@ -268,6 +309,24 @@ def get_apns_certificate(self, application_id=None):
...
+	def _get_apns_auth_key_path(self, application_id=None):
+		return self._get_application_settings(application_id, "APNS", "AUTH_KEY_PATH")
+
+	def _get_apns_auth_key_id(self, application_id=None):
+		return self._get_application_settings(application_id, "APNS", "AUTH_KEY_ID")
+
+	def _get_apns_team_id(self, application_id=None):
+		return self._get_application_settings(application_id, "APNS", "TEAM_ID")
+
+	def get_apns_token_lifetime(self, application_id=None):
```
jleclanche: This method/setting looks unused.

push_notifications/conf/legacy.py
```diff
@@ -14,6 +14,9 @@ class empty(object):
```
jleclanche: Why is this added to the legacy config? If we're no longer defaulting to it / recommending it I don't think it's worth adding more features to it.

jleclanche:
We can add djangomaster to allow_failures although I'm surprised rest framework isnt up to date yet.

So, at a first glance it looks good although I would have preferred better-split commits for review. Lots of style issues (please use lower_case_underscore for variables as pep8 recommends).

Left some comments for review. Needs work but I'd like to have it land.

push_notifications/models.py
shigmas: Sure. I guess if someone doesn't want to generate the JWT credentials on every single request, they use the PyAPNS API.

push_notifications/conf/app.py
shigmas: That's a mistake. The token is intended to be generated, and used for up to an hour. I did test it with PyAPNS2, but didn't finish propagating it up.

push_notifications/conf/legacy.py
shigmas: But I think it should still work. It's still default.

It's another discussion, but FCM and APNS support for multiple applications (one shared token) and the app config (no sharing) make it seem like another approach would be better, from a code and configuration standpoint.

shigmas:
Okay - I'm not sure how to get travis to ignore the failures. @jleclanche - what's the .travis.yml syntax that I'm screwing up?

jleclanche:
This is what you need: https://github.com/dj-stripe/dj-stripe/blob/master/.travis.yml#L41-L43

shigmas:
@jleclanche Legacy is still the default and needs to be supported. Any other changes before I squash them into one?

shigmas:
With PR's piling up, I don't want this one to get forgotten. Any more changes before I squash?

push_notifications/conf/legacy.py
matthewh: I agree with @jleclanche here. We want to get people using the new format and drop legacy. Legacy was provided as a way to keep existing configurations working.

shigmas: But shouldn't that be a separate change? It's still default now. Shouldn't it at least be deprecated first, and then we remove support?

matthewh: I'm not saying remove legacy. I'm saying only add the JWT support to the new config format. I guess if the work is already done though we should just move forward with it as is.

shigmas: It was trivial to support legacy. If a smaller change was desired, supporting the *new* config would have been a separate change.

jleclanche:
@shigmas Sorry, I guess it got forgotten :/ Can you rebase on top of master, squash the stylistic commits and get rid of the merge commits?

WRT legacy I'm not clear enough on how DPN works in that regard right now, so this is fine. I'm ok landing this once the commits get cleaned up.

chrisballinger:
@auvipy This can be closed now that #533 was merged

Issue #429, opened 2017-10-11
Support for `apns-collapsed-id`


ezheidtmann:
It appears that this package does not support the APNS feature `apns-collapse-id`. The apns2 library includes support as a kwarg to `send_notification` or `send_notification_batch` but that kwarg cannot be set via this package's `_apns_send()`.

I don't have a fix available, just documenting that this feature appears to be missing. Thanks.

ezheidtmann:
Whoops, I was looking at the wrong client library. Updated issue description.

roshanr95:
Any updates on this?

ezheidtmann:
@roshanr95 The PR is working for me in production. If you can run the tests & review my work, that could help get it included in a future release. Thanks.

alemangui:
Will this fix be included in an upcoming release? Would be very useful

matthewh:
This is fixed on master and will be present in the next release.

PR #430, opened 2017-10-11
Support `apns-collapse-id` via `collapse_id` kwarg

ezheidtmann:
Re #429 

This would appear to work but hasn't been tested. I can't figure out how to run existing tests. :upside_down_face:

### DIFF
--- a/push_notifications/apns.py
+++ b/push_notifications/apns.py
@@ -78,6 +78,8 @@ def _apns_send(
 		except ValueError:
 			raise APNSUnsupportedPriority("Unsupported priority %d" % (priority))
 
+	notification_kwargs["collapse_id"] = kwargs.pop("collapse_id", None)
+
 	if batch:
 		data = [apns2_client.Notification(
 			token=rid, payload=_apns_prepare(rid, alert, **kwargs)) for rid in registration_id]
--- a/tox.ini
+++ b/tox.ini
@@ -15,6 +15,7 @@ deps =
 	pytest
 	pytest-django
 	mock
+	pywebpush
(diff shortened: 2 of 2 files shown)


jamaalscarlett:
@ezheidtmann you run the tests using tox.  
```
pip install tox
```
then run using:
```
tox
```

ekimia:
what's the status of this?

ezheidtmann:
I'm running in prod and it appears to work. We are having trouble with local notifications *sometimes* not getting overwritten by our remote notification using the collapse id, but I believe that's a bug in our code not in the APNS code.

ezheidtmann:
I rebased on latest master. Looks like existing tests pass; Travis failures appear unrelated.

push_notifications/apns.py
```diff
@@ -78,6 +78,8 @@ def _apns_send(
 		except ValueError:
 			raise APNSUnsupportedPriority("Unsupported priority %d" % (priority))
 
+	notification_kwargs["collapse_id"] = kwargs.pop("collapse_id", None)
```
auvipy: this part needs to be tested to avoid any regression.

ezheidtmann:
Ok. Unfortunately I will not write this test any time soon.

sirdodger:
We have the same issue and are running off a private fork to fix the issue.

matthewh:
I've opened a new PR #523 and included a basic test to ensure the value is passed to apns2. If you were tracking this PR, can you please review #523 and tell me if that test is sufficient to accept the newer PR and close this?

PR #437, opened 2017-10-27
setup which works for fpm

### LINKED ISSUE #435: PIP package latest version
I need FCM notifications in my project, unfortunately 1.4.1 still doesn't have it.

I'm using fpm utility to build packages for my production.

`.gem/ruby/2.3.0/bin/fpm --python-pip /usr/bin/pip -fs python -t deb --no-python-downcase-name -p /tmp/out django-push-notifications`
this command runs `pip` with `--no-binary :all:` flag.
So I don't know why pip pulls out 1.4.1 version
```
# /usr/bin/pip download --no-clean --no-deps --no-binary :all: -i https://pypi.python.org/simple --build /tmp/package-python-build-7b309b186f34db30cf918a4697bf7cde1c58b1b833a84709691914ab122c/django-push-notifications django-push-notifications
Collecting django-push-notifications
  Using cached django-push-notifications-1.4.1.tar.gz
  Saved ./django-push-notifications-1.4.1.tar.gz
Successfully downloaded django-push-notifications
```

is it possible to make 1.5.0. make available to download this way, to build package using fpm (git repo doesn't work with `fpm` as well)?

theromis:
finally just created approximately appropriate setup args
please correct if I did something wrong

Fix #435

### DIFF
--- a/push_notifications/admin.py
+++ b/push_notifications/admin.py
@@ -14,7 +14,6 @@ class DeviceAdmin(admin.ModelAdmin):
 	list_filter = ("active",)
 	actions = ("send_message", "send_bulk_message", "enable", "disable")
 	raw_id_fields = ("user",)
-        readonly_fields = ('device_id', )
 
 	if hasattr(User, "USERNAME_FIELD"):
 		search_fields = ("name", "device_id", "user__%s" % (User.USERNAME_FIELD))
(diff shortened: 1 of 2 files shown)


matthewh:
Hello @theromis. Thank you for the pull request. This project moved away from defining the setup in setup.py and instead keeps everything in [setup.cfg](https://github.com/jleclanche/django-push-notifications/blob/master/setup.cfg). Do you know why FPM is having issues reading setup.cfg?

Also, for this PR to be accepted you need to ensure that the syntax issues are addressed.

jleclanche:
Run `pip install --upgrade pip setuptools wheel` in a fresh virtualenv before installing this. I've made the conscious decision of not supporting ancient versions of those and that's probably what you're hitting.

theromis:
@jleclanche  thank you, this solution works

PR #458, opened 2018-03-20
Swift 4: UIDevice.current.identifierForVendor

FahadAlbukhari:
Swift 4: UIDevice.current.identifierForVendor


jleclanche:
You'll need to update the most recent migration for that field as well, otherwise it'll complain about out-of-sync data.

jleclanche:
PS: Make sure to update the migration, not generate a new one.

PR #473, opened 2018-09-07
Avoid mutating arguments in `gcm._cm_send_request`

iharthi:
`gcm._cm_send_request` mutates its `data` argument by using `dict.pop`. Since this function may be called more than once from `push_notifications.GCMDeviceQuerySet.send_message`, with same dict as `data` argument, only the first call will receive proper values.

### DIFF
--- a/push_notifications/gcm.py
+++ b/push_notifications/gcm.py
@@ -124,10 +124,10 @@ def _cm_send_request(
 	if cloud_type == "FCM" and use_fcm_notifications:
 		notification_payload = {}
 		if "message" in data:
-			notification_payload["body"] = data.pop("message", None)
+			notification_payload["body"] = data["message"]
 
 		for key in FCM_NOTIFICATIONS_PAYLOAD_KEYS:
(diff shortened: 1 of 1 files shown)


push_notifications/gcm.py
```diff
@@ -124,10 +124,10 @@ def _cm_send_request(
 	if cloud_type == "FCM" and use_fcm_notifications:
 		notification_payload = {}
 		if "message" in data:
-			notification_payload["body"] = data.pop("message", None)
+			notification_payload["body"] = data["message"]
 
 		for key in FCM_NOTIFICATIONS_PAYLOAD_KEYS:
-			value_from_extra = data.pop(key, None)
+			value_from_extra = data[key] if key in data else None
```
jleclanche: data.get(key)

jleclanche:
Was there an actual reason these were getting popped before?

iharthi:
You're right, there is a reason, `data` is used later. Closing PR for now, will work on a more appropriate fix and maybe add tests for this particular issue, then recreate PR later.

PR #474, opened 2018-09-07
Fix gcm mutating extra args

iharthi:
`gcm._cm_send_request` mutates its `data` argument by calling `dict.pop` for all items in `FCM_NOTIFICATIONS_PAYLOAD_KEYS` and then using the rest as `payload["data"]`. However, since this function may be called more than once from `push_notifications.GCMDeviceQuerySet.send_message`, with same dict as `data` argument, only the first call will have extra values in `payload["notification"]`.

Suggested fix creates a copy of the argument before mutating it. A test case that reproduces the error if ran without the fix is also included.

### DIFF
--- a/push_notifications/gcm.py
+++ b/push_notifications/gcm.py
@@ -119,6 +119,8 @@ def _cm_send_request(
 
 	payload = {"registration_ids": registration_ids} if registration_ids else {}
 
+	data = data.copy()
+
 	# If using FCM, optionnally autodiscovers notification related keys
 	# https://firebase.google.com/docs/cloud-messaging/concept-options#notifications_and_data_messages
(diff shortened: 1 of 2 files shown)


jleclanche:
You didn't need to open a new PR, the old one can be updated with a force push.

push_notifications/gcm.py
```diff
@@ -119,6 +120,8 @@ def _cm_send_request(
 
 	payload = {"registration_ids": registration_ids} if registration_ids else {}
 
+	data = copy.copy(data)
```
jleclanche: There is a `dict.copy` method specifically for this. You don't need to import copy.

PR #492, opened 2019-03-12
Add support for python 3.7

odinuge:
Hi

Without this it is impossible to install using pip when running python 3.7.


odinuge:
Would be nice to get this into master and craft a release, so we can remove our patched fork. :smile:

PR #505, opened 2019-06-14
[WIP] Add deploy script to travis config

odinuge:


jezdez:
Thanks, this looks good in general, but I'd like to use [Travis build stages](https://docs.travis-ci.com/user/build-stages/) so deploys only happen when the whole test matrix has passed. I've opened #517 which contains the changes required which we can merge into master later to do the 2.0.0 release.

PR #507, opened 2019-06-25
Solve `Push failed: 403 Forbidden` problem with Chrome

wtayyeb:
Hi,
there was a hidden problem occur frequently when sending notification to WebPushDevice specially with Google Chrome! when you have several devices say some firefox and some chrome, if first notification sent to chrome device, there will no bug, but if the first notification sent to firefox device sending to any of chrome devices will fail. I have search the web but no clear solution for it found out there, that lead me to guess it is a library bug. I found that you pass settings.PUSH_NOTIFICATIONS_SETTINGS['WP_CLAIMS'] via `LegacyConfig` to `pywebpush.webpush`. the `webpush` method alter given vapid_claims and inject `aud`, `exp` into it **once**.
It is a bad practice to change such settings in this way at runtime. It also lead to a hidden bug that if first device is firefox, its `aud` will be firefox related url, and when sending to chrome device after that, it will rejected with `Push failed: 403 Forbidden`.

I try to solve this problem with minimum change to code. and hope you merge it as soon as possible. thanks.

### DIFF
--- a/push_notifications/webpush.py
+++ b/push_notifications/webpush.py
@@ -29,7 +29,7 @@ def webpush_send_message(
 			subscription_info=subscription_info,
 			data=message,
 			vapid_private_key=get_manager().get_wp_private_key(application_id),
-			vapid_claims=get_manager().get_wp_claims(application_id),
+			vapid_claims=get_manager().get_wp_claims(application_id).copy(),
 			**kwargs
 		)
 		results = {"results": [{}]}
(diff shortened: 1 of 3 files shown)


auvipy:
can you also add a unit test?

wtayyeb:
unittest? does it require it? I will try. but don't think have that expertise to do it right!

wtayyeb:
@auvipy there is no test case for webpush at all, create new one?

auvipy:
it would e great to have

wtayyeb:
@auvipy unittest added and passed.

PR #516, opened 2019-08-16
Update supported version of apns2

### LINKED ISSUE #513: unpinned dependencies (apns2)
apns2 released a 0.6.0 version that uses type hinting and as such won't work on Python versions older than 3.5.0

odinuge:
There are some breaking changes in `apns2===0.6.0`, so we have to limit the supported version. They also introduced typings, essentially dropping support for python 3.4 (that we support).

Fixes #513 

I think this, together with the fix to add support for python 3.7, should be cherry-picked into a new release `1.6.1`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ classifiers =
 [options]
 packages = find:
 install_requires =
-	apns2>=0.3.0
+	apns2>=0.3.0,<0.6.0
 	pywebpush>=1.3.0
 	Django>=1.11
 


setup.cfg
```diff
@@ -27,7 +27,7 @@ classifiers =
 [options]
 packages = find:
 install_requires =
-	apns2>=0.3.0
+	apns2>=0.3.0,<0.6.0
```
auvipy: 0.6.0 is the latest version, shouldn't this be >=0.6.0,<0.7?

odinuge: Hmm. Not sure if I understand you.

`0.6.0` is the version the doesn't work (because of the changes),  so using `>=0.6.0,<0.7` will imply that we support `0.6.0`, or am I wrong?

jezdez:
Hm, changing the base of the PR isn't sufficient for a clean merge into the new 1.6.x branch. I'll cherry-pick the commit of this PR manually.

jezdez:
Closing in favor of #517 which contains this fix.

PR #522, opened 2019-08-29
apns-collapse-id with tests

matthewh:


matthewh:
PR didn't land as expected. Will resubmit later.

PR #551, opened 2020-01-18
Test python 3.8 and Django 3.0 and align doc version specs

hamstap85:
We're trying to upgrade to Django 3.0 as soon as possible, so I'm going through all our dependencies and helping ensure they work on the latest version of everything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ build
 
 # tox
 .tox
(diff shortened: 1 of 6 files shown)


aaronn:
Does anyone have pypi access to update this?

cc @hamstap85 , @auvipy

Issue #567, opened 2020-03-25
KeyError at /admin/push_notifications/apnsdevice/ when bulk sending test notification


DataGreed:
I use p8 cert with django-push-notifications 2.0.0

Whenever I go to admin site and try to send bulk push notification I get following error:

```
Environment:


Request Method: POST
Request URL: http://127.0.0.1:8000/admin/push_notifications/apnsdevice/

Django Version: 2.2.7
Python Version: 3.7.3
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'rest_framework.authtoken',
 'adminsortable2',
 'push_notifications']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback:

File "myproject/.env/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "myproject/.env/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "myproject/.env/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "myproject/.env/lib/python3.7/site-packages/django/contrib/admin/options.py" in wrapper
  606.                 return self.admin_site.admin_view(view)(*args, **kwargs)

File "myproject/.env/lib/python3.7/site-packages/django/utils/decorators.py" in _wrapped_view
  142.                     response = view_func(request, *args, **kwargs)

File "myproject/.env/lib/python3.7/site-packages/django/views/decorators/cache.py" in _wrapped_view_func
  44.         response = view_func(request, *args, **kwargs)

File "myproject/.env/lib/python3.7/site-packages/django/contrib/admin/sites.py" in inner
  223.             return view(request, *args, **kwargs)

[...]

PR #566, opened 2020-03-25
Updated README: added info about APNS_AUTH_KEY_PATH, APNS_AUTH_KEY_ID, APNS_TEAM_ID

DataGreed:
Updated README: added info about APNS_AUTH_KEY_PATH, APNS_AUTH_KEY_ID, APNS_TEAM_ID settings for using Apple's p8 push certificates.

This addresses issues #455  and #527

PR #568, opened 2020-03-25
#567 Fixes crash on bulk_send test message operation from admin

DataGreed:
the return format has changed for some reason, I adapted the bulk send action from admin site to not crash on send.

Fixes issue #568


DataGreed:
Thanks for merging 🙏

DataGreed:
@auvipy I think it would be a good idea to merge https://github.com/jazzband/django-push-notifications/pull/566 and create a new release so PIP will have a version that does not crash on bulk sending test messages

PR #569, opened 2020-03-31
Update rest_framework.py

jameslee961018:


push_notifications/api/rest_framework.py
```diff
@@ -28,19 +28,6 @@ def to_representation(self, value):
 		return value
 
 
-# Serializers
```
auvipy, line 31: why you deleted this part?

jameslee961018: I thought this part didn't need for your prodject.
I am  a senior Ptthon/ Django developer.

PR #572, opened 2020-04-22
Edit old migrations to use new USER_MODEL setting value

ezheidtmann:
The `USER_MODEL` setting appears very useful, but in the 2.0.0 release, the original tables are created with `user` as a ForeignKey to `settings.AUTH_USER_MODEL`.

This creates several problems, including that `./manage.py makemigrations` will create a migration in `push_notifications` that can't be committed or deployed if the project was installed from pypi.

This PR edits old migrations to use `USER_MODEL` directly. For new users, this seems to honor the intention of the feature. For existing users, they should expect trouble if they ever change the `USER_MODEL` or `AUTH_USER_MODEL` settings, with or without this patch.

### DIFF
--- a/push_notifications/migrations/0001_initial.py
+++ b/push_notifications/migrations/0001_initial.py
@@ -4,6 +4,8 @@ from django.db import migrations, models
 
 import push_notifications.fields
 
+from ..settings import PUSH_NOTIFICATIONS_SETTINGS as SETTINGS
+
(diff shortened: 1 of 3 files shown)


ezheidtmann:
Rebased on the latest release `3.0.0`

joeyorlando:
bump

PR #599, opened 2021-01-29
Drop python 2.x and django before 2.2

sevdog:
Since Python 2.x is EOL and also Django before 2.2 tests and code are updated to work with python3.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
-	py{35,36,37,38,39}-django{111,20,21,22}-apns{030,060}
-	py{36,37,38,39}-django30-apns{030,060}
+	py{36,37,38,39}-django{22}-apns{030,060}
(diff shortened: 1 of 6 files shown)


auvipy (approved):

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
-	py{35,36,37,38,39}-django{111,20,21,22}-apns{030,060}
-	py{36,37,38,39}-django30-apns{030,060}
+	py{35,36,37,38,39}-django{22}-apns{030,060}
```
smithdc1, line 3: Jazzbands general approach has been to drop support for Python 3.5.

sevdog: That is not a problem. Just updated without python 3.5.

ZeroCoolHacker (approved):

auvipy (approved):

PR #600, opened 2021-02-15
Make push optional dependecies

sevdog:
In some use-cases this package is used only to serve a subset of available push services.

In this case having every push-library as a required dependecy may cause problems with the environment (ie: dependecies of dependency).

This PR turns APNS and webpush as optional dependencies using [`extras_require`](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies) option.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,8 +34,9 @@ Dependencies
 - Python 3.6+
 - Django 2.2+
 - For the API module, Django REST Framework 3.7+ is required.
-- For WebPush (WP), pywebpush 1.3.0+ is required. py-vapid 1.3.0+ is required for generating the WebPush private key; however this
+- For WebPush (WP), pywebpush 1.3.0+ is required (optional). py-vapid 1.3.0+ is required for generating the WebPush private key; however this
   step does not need to occur on the application server.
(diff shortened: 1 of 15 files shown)


auvipy (approved):

sevdog:
I have rabased this after #599.

sevdog:
I see that now travis is failing due to rust dependency ofcryptography...

sevdog:
@auvipy maybe the issue with crypthography on travis is due to [this change](https://cryptography.io/en/latest/changelog.html#v3-4): from version 3.4, released on Feb this year, it requires RUST to compile binaries. There is a [binary package](https://cryptography.io/en/latest/installation.html#rust) for ubuntu, but it requires pip to be updated, on travis PIP uses version 20.1.1 (May 2020) which may not get the updated binary. This problem maybe could be fixed by forcing PIP to update.

sevdog:
I have just seen that now this repo uses Gihub Action instead of Travis. I will update this PR as soon as I can.

Andrew-Chen-Wang (approved):
LGTM @auvipy anything else?

auvipy (approved):

PR #603, opened 2021-03-12
Update settings.py

Fethienv:
Add WP_TTL option


auvipy (changes requested):
cant you add this to the WP_TTL adding PR?

jamaalscarlett:

PR #605, opened 2021-03-12
add WP_TTL

Fethienv:
add WP_TTL  for store message till user open browser


jamaalscarlett:

PR #614, opened 2021-07-06
Spelling error & convert string to be translatable

Andrew-Chen-Wang:


auvipy:
migration was missing in this PR

PR #615, opened 2021-07-08
Generic FCM device convenience model

### LINKED ISSUE #618: Switch over from GCM to FCM
My understanding that nobody uses GCM anymore, so perhaps it is time for the upgrade?

My suggestion would be to rename GCM model to FCM model (breaking change) and add `platform` choices/enum to it, like I've done in the https://github.com/jazzband/django-push-notifications/pull/615

I agree perhaps there is no reason to keep the GCM model anymore and `message_type` but it will be a (small) breaking change.

What are your thoughts?

1. Rename GCM to FCM?

2. Keep GCM as a proxy model for backward compatibility? (seems like unnecessary work)

3. platform choices: 

a) android, ios, web - full text in the database consistent with the rest of the library
b) a, i, w - save space
c) 0, 1, 2 - save space, easy indexing, sorting
d) new django enums

ayushin:
Hi all,

As FCM can now handle ios/android/web we found it convenient to have a generic FCMDevice model that also keep the data about device platform.

Here is my take on how this could be implemented, any feed back is very welcome.


push_notifications/migrations/0008_fcmdevice.py
```diff
@@ -0,0 +1,25 @@
...
+    dependencies = [
+        ('push_notifications', '0007_uniquesetting'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='FCMDevice',
+            fields=[
+                ('gcmdevice_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='push_notifications.gcmdevice')),
+                ('platform', models.CharField(blank=True, choices=[('i', 'ios'), ('a', 'android'), ('w', 'web')], help_text='Optional device platform: i - ios, a - android, w - web', max_length=1, null=True)),
```
Andrew-Chen-Wang, line 18: I'd much rather see a small positive integer field for faster filtering on the database.

push_notifications/models.py
```diff
@@ -257,3 +256,29 @@ def send_message(self, message, **kwargs):
 		return webpush_send_message(
 			uri=self.registration_id, message=message, browser=self.browser,
 			auth=self.auth, p256dh=self.p256dh, application_id=self.application_id, **kwargs)
+
+
+class FCMDevice(GCMDevice):
+	PLATFORM_IOS = 'i'
```
Andrew-Chen-Wang, line 262: doesn't seem to be indented properly

Andrew-Chen-Wang (changes requested):
I don't understand this. The GCM module currently uses Firebase, so why add a new model?

ayushin:
Convenience to use FCM by default plus type of device

Andrew-Chen-Wang:
@ayushin got it, understood. I don't think this PR is necessary then since devs should override the GCM model. Plus the name just makes it confusing since the current model is still called GCM even though GCM doesn't exist anymore.

ayushin:
Yeah well, perhaps we can just turn around and rename GCM to be FCM (I think it is useful to have a platform field there) and GCM to be a proxy model to FCM for backward compatibility?

push_notifications/migrations/0008_fcmdevice.py
ayushin: That sounds like a good idea indeed.

push_notifications/models.py
Andrew-Chen-Wang, line 262: Still not indented properly. 4 chars.

push_notifications/models.py
```diff
@@ -257,3 +256,29 @@ def send_message(self, message, **kwargs):
 		return webpush_send_message(
 			uri=self.registration_id, message=message, browser=self.browser,
 			auth=self.auth, p256dh=self.p256dh, application_id=self.application_id, **kwargs)
+
+
+class FCMDevice(GCMDevice):
```
Andrew-Chen-Wang, line 261: The migration for existing users doesn't make sense. You should instead be renaming the current model to GCP no?

Andrew-Chen-Wang (changes requested):

Andrew-Chen-Wang:
One more problem. This currently still uses FCM legacy API rather than v1. Please visit django fcm for how I migrated it. At some point, the legacy API will be deprecated and gone.

push_notifications/api/rest_framework.py
```diff
@@ -110,6 +110,12 @@ def validate_device_id(self, value):
 		return value
 
 
+class FCMDeviceSerializer(GCMDeviceSerializer, ModelSerializer):
+	class Meta(GCMDeviceSerializer.Meta):
+		model = FCMDevice
+		fields = GCMDeviceSerializer.Meta.fields + ('platform',)
```
jamaalscarlett, line 116: Please see https://github.com/jazzband/django-push-notifications/blob/master/CONTRIBUTING.md regarding single quotes

jamaalscarlett (changes requested):

ayushin:
Would it be better to replace the GCMModel with FCMModel instead? We'll have a cleaner code and have a migration renaming one in the previous installs

jamaalscarlett:
@ayushin,  I agree with @Andrew-Chen-Wang that there is currently more work needed to properly support the v1 FCM implementation.  I do agree that a new FCMDevice model makes sense, but it should be using the firebase_admin sdk

ayushin:
Agreed. Actually https://github.com/xtrinch/fcm-django makes more sense for FCM

Andrew-Chen-Wang:
@ayushin I made the migration for fcm-django. FCM-django is just a fork of django push notifications before this package added fcm support I believe.

In other words, you can basically copy my code and add it to this package.

jamaalscarlett:
@Andrew-Chen-Wang Do you want to do it?  You did the work, you should get the credit :)

Andrew-Chen-Wang:
Really squeezed for time lately. Will loop back around in a month maybe to implement if necessary.

jamaalscarlett:
I get it.  I will be on PTO all next week, but maybe I will take a crack at it after that

PR #623, opened 2021-10-07
Add Python 3.10 support and classifiers

Andrew-Chen-Wang:


bertonha:
if this branch is rebased with latest master it should pass

Andrew-Chen-Wang:
We'll need to drop hyper for Python 3.10 support, which probably comes from pywebpush

bertonha:
> We'll need to drop hyper for Python 3.10 support, which probably comes from pywebpush

hyper comes from apns2

Issue #559, opened 2020-02-14
Passing the image field in options


cychongaspevo:
How  to pass in image data like the example show in  this page for FCM.

https://firebase.google.com/docs/cloud-messaging/ios/send-image

ashishnitinpatil:
This PR fixed the issue but apparently it was closed - https://github.com/jazzband/django-push-notifications/pull/619

Will try to make a PR to fix the problem, so that you could simply do

```
device.send_message(..., image="https://www.example.com/image.png")
```

PR #624, opened 2021-10-07
++Enable sending image in GCM notifications

ashishnitinpatil:
Existing behavior sends it under the `data` field which doesn't behave correctly with the native image notifications.

Fixes https://github.com/jazzband/django-push-notifications/issues/559

### DIFF
--- a/push_notifications/gcm.py
+++ b/push_notifications/gcm.py
@@ -25,7 +25,7 @@ FCM_OPTIONS_KEYS = [
 	"restricted_package_name", "dry_run", "mutable_content"
 ]
 FCM_NOTIFICATIONS_PAYLOAD_KEYS = [
-	"title", "body", "icon", "sound", "badge", "color", "tag", "click_action",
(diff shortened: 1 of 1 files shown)


LeeHanYeong:
To deliver an image to the recipient in FCM's notifications payload, you must use the "image" key.
I hope there is a quick release.

https://firebase.google.com/docs/cloud-messaging/android/send-image

alexislg2:
Thanks, it is so easy. Waiting for the relase!

astrike30:
Can anyone confirm that this is working? I have tried passing `image=url...` and as `extra={"image": url}` but I cant seem to get either to work.

ashishnitinpatil:
Yes @astrike30 I can confirm it works with `extra={"image": url}` where `url` is well-formed & publicly accessible. Sometimes, the FCM network fails showing the image for some random reasons, so make sure to test on multiple devices if one doesn't seem to work.

PR #631, opened 2021-11-12
Add Edge support to webpush

simonkern:
This adds Edge support to webpush and showcases an easier way to extract browsername and version for Chrome, Edge, and Opera in the README.

Any feedback would be greatly appreciated.


auvipy (approved):

Andrew-Chen-Wang:
@simonkern I don't know if this statement is correct, but the edge url seems to change all the time? Is it possible to make this an EDGE setting?

push_notifications/models.py
```diff
@@ -240,7 +241,7 @@ class WebPushDevice(Device):
 	browser = models.CharField(
 		verbose_name=_("Browser"), max_length=10,
 		choices=BROWSER_TYPES, default=BROWSER_TYPES[0][0],
-		help_text=_("Currently only support to Chrome, Firefox and Opera browsers")
+		help_text=_("Currently only support to Chrome, Firefox, Edge and Opera browsers")
```
Andrew-Chen-Wang, line 244: Please create a migration!

Andrew-Chen-Wang (changes requested):

simonkern:
> @simonkern I don't know if this statement is correct, but the edge url seems to change all the time? Is it possible to make this an EDGE setting?

I was not able to produce an URL with another domain so far. However, I was also unable to find proper documentation on this matter. But if this URL would change all the time, it would be necessary to adapt the model to store the baseURL as well, imho. What do you think and suggest?

Andrew-Chen-Wang:
I'm not sure what you mean by adapt the model to store the URL. I was suggesting in settings.py, instead of hard coding the url, allow for user customization like the chrome url. Additionally, make sure to create a migration!

simonkern:
Doesn't WP_POST_URL already solve this?

Andrew-Chen-Wang (approved):
ah gotcha sorry 😅 LGTM then

Andrew-Chen-Wang:
AHH but wait migration file!

Andrew-Chen-Wang:
Thanks again!

PR #632, opened 2021-11-14
pass alert parameter as a function which accepts token parameter in o…

Phishion:
pass alert parameter as a function which accepts token parameter in order to set different message per user


auvipy (changes requested):
please fix the failing tests

PR #634, opened 2021-11-14
Different messages be specified for each user

Phishion:
pass alert parameter as a function which accepts token parameter in order to set different message per user

### DIFF
--- a/push_notifications/apns.py
+++ b/push_notifications/apns.py
@@ -42,20 +42,23 @@ def _apns_prepare(
 	token, alert, application_id=None, badge=None, sound=None, category=None,
 	content_available=False, action_loc_key=None, loc_key=None, loc_args=[],
 	extra={}, mutable_content=False, thread_id=None, url_args=None):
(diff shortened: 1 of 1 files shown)


jamaalscarlett:
@Phishion This looks good, but it appears that the indentation is off.  And can you update the readme with an example like the badge?

jamaalscarlett (changes requested):
I think this needs to be indented

Phishion:
> I think this needs to be indented

Hello, I have updated the indentation and added the [description](https://github.com/jazzband/django-push-notifications/pull/634/commits/b19e8816773c7683a3f33e1f3b738ce68eeb7c5a) for readme, just contact me if you have any questions

jamaalscarlett:
@Phishion Can you add the description to the README.rst file?

Phishion:
> @Phishion Can you add the description to the README.rst file?

The latest PR is [here](https://github.com/jazzband/django-push-notifications/pull/653), this one will be closed.

PR #639, opened 2021-12-10
chore: test only active Django versions

bertonha:
Django 3.0 and 3.1 already got on EOF

I also had forgot to add the Django 4.0 on GHA


bertonha:
would make sense to remove python 3.6 and 3.7 since django 4.0 does not support it anymore ?

https://docs.djangoproject.com/en/4.0/releases/4.0/#python-compatibility

auvipy (approved):

Issue #640, opened 2021-12-13
Incompatibility with userAgentData.


dismine:
Hi.

I used ``loadVersionBrowser`` function from the documentation. And my view stopped working without producing any errors in logs. Later I figured out that it was caused by unexpected browser name. I use Vivaldi and  ``userAgentData`` recognized it as __Chromium__, not __Chrome__ as we expect, while old user agent string still claims that we are Chrome.

simonkern:
The linked PR could also solve your issue. Could you try it?

dismine:
Hi.

I have tested your code. For Vivaldi the situation is the same. ``navigator.userAgentData`` returns ``Chromium``. It correctly returns ``Chrome`` only when fallback to parsing a user agent string. In my case it is ``Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.113 Safari/537.36``.

So, does it solve my issue? Kind of, yes. But why modern way to get a browser brand doesn't work with Chromium based browsers?

simonkern:
```if (brand.match(/opera|chrome|edge|safari|firefox|msie|trident/i) !== null)```

This condition can never be true for the string "Chromium", which is why it should fallback to userAgent anyway. Are you sure the function returns "Chromium" as browsername?

```
"Chromium".match(/opera|chrome|edge|safari|firefox|msie|trident/i)
null
```

dismine:
> Are you sure the function returns "Chromium" as browsername?

Your function returns ``Chrome``. So, yes, it works correctly.

> Vivaldi wouldn't be in the regexp anyway. Why they do not supply their browsername, like Edge does - I have no idea...

I can only [guess](https://vivaldi.com/blog/user-agent-changes/).

jamaalscarlett:
@dismine https://github.com/jazzband/django-push-notifications/pull/643 seems to have fixed your issue, can we close this?

PR #643, opened 2021-12-30
More robust handling of userAgentData, fix WebPushDeviceAdmin

simonkern:
Relevant userAgentData ist not always stored in navigator.userAgentData.brands[0]

It is necessary to loop through all available navigator.userAgentData.brands and try to find the relevant entry. In case of Edge (and maybe in case of other chromium-based browsers too) it is also necessary to ignore any matches regarding chrome...

addresses #640

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -211,8 +211,8 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
 			const browserversion = browser.version;
 
 			return {
-			    name: browsername,
-			    version: browserversion
+				name: browsername,
+				version: browserversion
 			}
 		}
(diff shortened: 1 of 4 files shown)


simonkern:
Field "device_id" is not available for WebPushDevices. Instead I added "browser" to ListView and "registration_id" to search_fields

auvipy (changes requested):
can you provide some tests for the changes please?

simonkern:
@auvipy what do you have in mind? Except for the field changes in the broken admin, these are only changes to the README.

auvipy:
@dismine

dismine:
New ``function loadVersionBrowser`` works for me. Already ported to my website.

auvipy (approved):

PR #648, opened 2022-01-24
feat: Add python3.10 support (ugly workaround)

bertonha:
the sub dependency of apns2 `hyper` is completely abandoned and it does not support python 3.10
simple because it uses some packages from `collections` where on python 3.10 it moved to `collections.abc`

I think the clean way to properly fix it, would be to move to a more maintained package to do Apple Push Notifications.
which might be [aioapns](https://github.com/Fatal1ty/aioapns)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


Andrew-Chen-Wang:
apns2 is a pain since no one maintains this valuable package. We should get everyone to use a fork instead.

PR #487, opened 2018-12-20
Improve Webpush, registration id unique configurable and details

goinnn:
Hi!

I'm sorry for these 6 mixed commits, although I think it is a nice christmas present still from my mate @rubgombar1 and me :-)

We use django-push-notifications from 3 years ago and we are very grateful and interested in improve this increible Django app.

For this reason, 1 year ago we introduce Web Push feature, and now we have several changes for it. [b6935832fc1009e2cbba7e25128282bb754191d6][35222c0fd349e6a520da31ec196035f5bbdddb18]:

- If you wanted send a push for three devices and one was outdated, currently the bulk process did not send to the next devices.
- If a devices was outdated, it was not disabled.

Also, now registration_id field depends on model (push type) to define if it is unique or not. We have developed a new feature, and now it depends on the settings.[6da01a12883f66e67e22b85744cf542a38f48dea][68d244ce955dde9c59316d9630d1b26c25fe9bc9]

Also, we have changes cleaning code [1f526069a45d82ff873156dedf9c57868969b875] and fixing a silly error sending a message from admin [f4b8ebfa181cc7bc5c126396cbd44c9aee6f5a95]

And a new commit to fix tests [b38b5ff1e8e091ddcd0f72f56ab9df5f91279545] 

These commits were in our production enviroment from:

1. 2018-07-23: [1f526069a45d82ff873156dedf9c57868969b875][f4b8ebfa181cc7bc5c126396cbd44c9aee6f5a95][b6935832fc1009e2cbba7e25128282bb754191d6][6da01a12883f66e67e22b85744cf542a38f48dea][68d244ce955dde9c59316d9630d1b26c25fe9bc9]  
2. 2018-09-24: [35222c0fd349e6a520da31ec196035f5bbdddb18] 
3. it is a new commit: [b38b5ff1e8e091ddcd0f72f56ab9df5f91279545], but it is only for fix pep8 & pyflakes errors

So, these features were very tested, but we don't have had any time to do a PR until now. But, of course we understand that you want see/analyze these features calmly.

Issue #673, opened 2022-11-04
[WebPush] Support for Safari


blighj:
Safari 16.1 on Mac OS Ventura supports WebPush, with the potential for it to come to iOS in 2023

It will need a code change to make it work with django-push-notifications.

As I understand it, when you register for a push notifaction on the client side, the browser gives you the full endpoint url to use. The way it is modeled in the code here, is that we only want the last part of the path from the endpoint url (some sort of token) and will determine the full endpoint url based on pre configured settings for each browser. These settings don't exist for Safari yet.

Apple and Microsoft seem to have made the same implementation decision where the push notifaction endpoints can actually be one of multiple domains, (*.notify.windows.com, *.push.apple.com). So fixing a browser to just one endpoint may not be the best way to model the standard.

Would it be perferable to store the full endpoint url given by the browser in the registration_id of the model, instead of just the token and use exactly what the browser gave us when we send the message. We could depercate the old way, but still support it for a while with a fallback. 
https://github.com/jazzband/django-push-notifications/blob/a20819258cc33ed7855db7231831bb9231dfe1ca/push_notifications/webpush.py#L7
```diff
def get_subscription_info(application_id, uri, browser, auth, p256dh):
-	url = get_manager().get_wp_post_url(application_id, browser)
+	if uri.startswith("https://"):
+		endpoint = uri
+	else:
+		url = get_manager().get_wp_post_url(application_id, browser)
+		endpoint = "{}/{}".format(url, uri)	
	return {
-		"endpoint": "{}/{}".format(url, uri),
+		"endpoint": endpoint,
		"keys": {
			"auth": auth,
			"p256dh": p256dh,
		}
	}
```

If people are happy with that general approach I can do up a PR with updated docs and tests.

viktorisacenko:
Adding Safari to BROWSER_TYPES in models should work

```python
BROWSER_TYPES = (
	("CHROME", "Chrome"),
	("FIREFOX", "Firefox"),
	("OPERA", "Opera"),
	("EDGE", "Edge"),
        ("SAFARI", "Safari"). <--------
)
```

viktorisacenko:
If you want to add webpush for Safari 16+ you need only to add the Browser type for safari in the models

push_notifications.models.py
```python
BROWSER_TYPES = (
	("CHROME", "Chrome"),
	("FIREFOX", "Firefox"),
	("OPERA", "Opera"),
	("EDGE", "Edge"),
[...]

Issue #665, opened 2022-07-17
How to setup properly just for web?


saileshkush95:
I want to setup push notification for website. I just follwed as described in documenation but Unfurtunately it is now working.

Here is my settings.
```
PUSH_NOTIFICATIONS_SETTINGS = {
    "WP_PRIVATE_KEY": BASE_DIR / 'private_key.perm',
    "WP_CLAIMS": {'sub': "mailto: development@example.com"}
}
```
and here is my client side code

```
<!doctype html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport"
          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>Push Notifications</title>
</head>
<body>
<h1>Push Notification Example</h1>
<script>
  // Utils functions:
  function urlBase64ToUint8Array(base64String) {
    var padding = '='.repeat((4 - base64String.length % 4) % 4)
    var base64 = (base64String + padding)
      .replace(/\-/g, '+')
      .replace(/_/g, '/')

    var rawData = window.atob(base64)
    var outputArray = new Uint8Array(rawData.length)

    for (var i = 0; i < rawData.length; ++i) {
      outputArray[i] = rawData.charCodeAt(i)
    }
    return outputArray;
  }

  function loadVersionBrowser(userAgent) {
    var ua = userAgent;
    var tem;
    var M = ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || [];
    if (/trident/i.test(M[1])) {
      tem = /\brv[ :]+(\d+)/g.exec(ua) || [];
      return {name: 'IE', version: (tem[1] || '')};
    }
    if (M[1] === 'Chrome') {
      tem = ua.match(/\bOPR\/(\d+)/);
      if (tem != null) {
        return {name: 'Opera', version: tem[1]};
      }
    }
    M = M[2] ? [M[1], M[2]] : [navigator.appName, navigator.appVersion, '-?'];
    if ((tem = ua.match(/version\/(\d+)/i)) != null) {
      M.splice(1, 1, tem[1]);
    }
    return {
      name: M[0],
      version: M[1]
    };
  };
[...]

Issue #658, opened 2022-05-29
requestPOSTToServer is not defined


Alzabel999999:
In your js cannot find function requestPOSTToServer. Where I can get it?

viktorisacenko:
#665

blighj:
@Alzabel999999
Any feedback on this PR, #674, specfically this file, which aims to explain wepush configuration
https://github.com/jazzband/django-push-notifications/blob/4fb64be08a5e9d4b9275e9632426d022d5a84a70/docs/WebPush.rst

azmeuk:
Until #674 is merged, you can find an implementation in the README:

https://github.com/jazzband/django-push-notifications/blob/bd387e954318469af227a2a6cb89c225c4ca5092/README.rst?plain=1#L383-L396

PR #674, opened 2022-11-20
Add WebPush support for Safari

blighj:
PR for #673
Stores the full endpoint url for web push given by the browser, in the registration_id of the model, instead of just the token, and use exactly what the browser gave us when we send the message. Still supports existing registration_id/browser combo's but rasies a warning.

Also:
 - Adds tests for web push
 - Update docs for configuring web push
 - Handle 404/410 when regeistering and deactivate device
 - Remove pywebpush from APNS' options.extras_require in setup.cfg

The new docs address issues/questions raised in #647, #658, #665 
The PR was inspired/informed by #487, #558, #602, #603, #604, #605


simonkern:
> tores the full endpoint url for web push given by the browser, in the registration_id of the model, instead of just the token, and use exactly what the browser gave us when we send the message.

I also thought about this aspect some time ago when I added push support and think this is a really good idea. This could also remove the need to extract the browser type via javascript, because we then already have the url of the endpoint we need to send the push to stored in the WebPushDevice object.

simonkern:
iOS 16.4 just landed. What's the current status here? Are there still things that need to be done?

blighj:
From my prespective, as a first time contributor, this is ready to go.
I assume the next step is a code review.

simonkern:
I am not a maintainer, but I could take this for a test spin. Do you know who could do a release?

push_notifications/webpush.py
```diff
@@ -1,39 +1,55 @@
...
+		results = {"results": [{"original_registration_id": device.registration_id}]}
+		if response.ok:
 			results["success"] = 1
+		else:
+			results["failure"] = 1
+			results["results"][0]["error"] = response.content
 		return results
 	except WebPushException as e:
+		if e.response is not None and e.response.status_code in [404, 410]:
+			results["failure"] = 1
```
martinrlilja, line 50: I get an error here because the exception comes from line 34, but `results` is defined on line 41. It should be enough to just switch them around.

ammarjmhor10:
Hi, is there any update please?

push_notifications/webpush.py
blighj: Excellent point, commit added to address it.

mamal72:
Hey guys. Any updates on this? Can we do one more push and have it released? Thanks in advance.

push_notifications/webpush.py
```diff
@@ -1,39 +1,55 @@
...
 
 def get_subscription_info(application_id, uri, browser, auth, p256dh):
-	url = get_manager().get_wp_post_url(application_id, browser)
+	if uri.startswith("https://"):
+		endpoint = uri
+	else:
+		url = get_manager().get_wp_post_url(application_id, browser)
+		endpoint = "{}/{}".format(url, uri)
+		warnings.warn(
+			"registratipn_id should be the full endpoint returned from pushManager.subscribe",
```
cuylerstuwe, line 16: Typo:

```suggestion
			"registration_id should be the full endpoint returned from pushManager.subscribe",
```

setup.cfg
```diff
@@ -38,7 +38,6 @@ setup_requires =
 APNS =
 	apns2>=0.3.0
 	importlib-metadata;python_version < "3.8"
-	pywebpush>=1.3.0
 	Django>=2.2
```
jamaalscarlett, line 41: Wont this throw an import error? You are still importing pywebpush in webpush.py

docs/WebPush.rst
```diff
@@ -0,0 +1,225 @@
...
+	    fetch(
+	      '<your endpoint url>',
+	      requestOptions
+	    )
+	  ).then((response) => response.json())
+	}
+
+Server Side logic to create webpush
+------------------------------
+Is is up to you how to add a view in your django application that can handle a POST of p256dh, auth, registration_id and create a WebPushDevice with those values assoicated with the appropriate user.
```
azmeuk, line 124: Typo: *Is is* -> *It is*

docs/WebPush.rst
```diff
@@ -0,0 +1,225 @@
+At a high-level, the key steps for implementing web push notifications after installing django-push-notifications[WP] are:
```
azmeuk, line 1: Thank you very much for the documentation!

README.rst
```diff
@@ -122,238 +122,11 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
...
-
-	Application Server Key = BEFuGfKKEFp-kEBMxAIw7ng8HeH_QwnH5_h55ijKD4FRvgdJU1GVlDo8K5U5ak4cMZdQTUJlkA34llWF0xHya70
-
-
-- Configure settings:
-
 - ``WP_PRIVATE_KEY``: Absolute path to your private certificate file: os.path.join(BASE_DIR, "private_key.pem")
 - ``WP_CLAIMS``: Dictionary with the same sub info like claims file: {'sub': "mailto: development@example.com"}
 - ``WP_ERROR_TIMEOUT``: The timeout on WebPush POSTs. (Optional)
-- ``WP_POST_URL``: A dictionary (key per browser supported) with the full url that webpush notifications will be POSTed to. (Optional)
```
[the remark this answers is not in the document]
blighj: This MR is essentially deprectating the old approach. Instead the full endpoint that the browser returns is stored in the registration_id of the model, instead of just the token, and the code uses exactly what the browser gave us when we send the message. Rather than composing a url based on what is in WP_POST_URL.

As a result I thought it was sensible to remove the WP_POST_URL from the docs as it would confuse new users.

docs/WebPush.rst
```diff
@@ -0,0 +1,225 @@
...
+
+These are in addition to the instalation steps for django-push-notifications[WP]
+
+Configure the VAPID keys
+------------------------------
+- Install:
+
+.. code-block:: python
+
+	pip install py-vapid  (Only for generating key)
```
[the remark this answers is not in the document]
blighj: This part of the docs is explaining how to generate the keys needed. It's my undestanding that you only need py-vapid for that part.
You could essentially do that part on your local machine, and then deploy your keys to production where the library is used in your application.

docs/WebPush.rst
blighj, line 124:
```suggestion
It is up to you how to add a view in your django application that can handle a POST of p256dh, auth, registration_id and create a WebPushDevice with those values assoicated with the appropriate user.
```

setup.cfg
blighj: It's my understanding on this one, that this set of requierments is for use with apples APNS, the WebPush(WP) section is lower down  and the pywebpush is declared there.
This line is a mistake in my eyes.

push_notifications/webpush.py
```diff
@@ -1,39 +1,55 @@
...
 		}
 	}
 
 
-def webpush_send_message(
-	uri, message, browser, auth, p256dh, application_id=None, **kwargs
-):
-	subscription_info = get_subscription_info(application_id, uri, browser, auth, p256dh)
-
+def webpush_send_message(device, message, **kwargs):
```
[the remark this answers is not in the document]
blighj: I see your point, but are you explictly asking me to change it back?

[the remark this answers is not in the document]
jamaalscarlett: I've thought about it some more.  I think it's fine.

push_notifications/webpush.py
jamaalscarlett: Could you add a comment explaining the two status codes and why you are disabling the device

push_notifications/webpush.py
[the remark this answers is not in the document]
azmeuk: I think this is less cumbersome to directly use WebpushDevice objects. However I understand your point about debugging @jamaalscarlett. I have no strong opinion, either way is fine to me.

azmeuk:
I tested this branch on several browsers, including Safari on iOS and I confirm this works!

azmeuk:
For what it worth, instead of manually using py-vapid, I could do the key generation dynamically with:

```python
from py_vapid import b64urlencode
from py_vapid import Vapid02 as Vapid
from cryptography.hazmat.primitives import serialization
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.asymmetric import ec


def vapid_public_key(path):
    vapid = Vapid.from_file(path)
    pk = vapid.public_key.public_bytes(
        serialization.Encoding.X962, serialization.PublicFormat.UncompressedPoint
    )
    return b64urlencode(pk)


def generate_private_key(path):
    private_key = ec.generate_private_key(ec.SECP256R1, default_backend())
    private_pem = private_key.private_bytes(
        encoding=serialization.Encoding.PEM,
        format=serialization.PrivateFormat.PKCS8,
        encryption_algorithm=serialization.NoEncryption()
    )
    with open(path, "wb") as file:
        file.write(private_pem)
        file.close()
```

`generate_private_key` can be used if the private key does not exist, `vapid_public_key` can be used to get the base64 export of the public key and insert it dynamically in the templates.

jamaalscarlett:
@blighj once you address @azmeuk I think this is good to merge.  Nice work!

push_notifications/webpush.py
blighj: how about?
```suggestion
			# if we find that the status from the vendor was 404 (invalide url) or 410 (Gone) we should no
			# longer try to send messages to this device.
			# https://web.dev/push-notifications-common-issues-and-reporting-bugs/#http-status-codes
			results["failure"] = 1
```

README.rst
```diff
@@ -122,238 +122,11 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
...
-
-	vapid --applicationServerKey
-
-	Application Server Key = BEFuGfKKEFp-kEBMxAIw7ng8HeH_QwnH5_h55ijKD4FRvgdJU1GVlDo8K5U5ak4cMZdQTUJlkA34llWF0xHya70
-
-
-- Configure settings:
-
 - ``WP_PRIVATE_KEY``: Absolute path to your private certificate file: os.path.join(BASE_DIR, "private_key.pem")
 - ``WP_CLAIMS``: Dictionary with the same sub info like claims file: {'sub': "mailto: development@example.com"}
```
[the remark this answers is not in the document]
blighj: It feels like more of an error in the docs, it's an invalid [mailto uri](https://datatracker.ietf.org/doc/html/rfc6068#page-4)
Rather than explain to people not to do that, I think it would be enough to corret the mistake in the docs
If you agree there is another place earlier in the doc that I'm aware of that has the same issue.

While we are here, this sentence doesn't make sense to me
>Dictionary with the same sub info like claims file:

I don't know what `the same sub info like claime file` is referring to
```suggestion
- ``WP_CLAIMS``: Dictionary with default value for the sub, (subject), sent to the webpush service, This would be used by the service if they needed to reach out to you, the sender. Could be a url or mailto e.g. {'sub': "mailto:development@example.com"}.
```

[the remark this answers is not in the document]
azmeuk: There is indeed another space after `mailto:` in a configuration example earlier in the doc. And there are occurences in `tests/settings_unique.py` and `tests/settings.py`, but this is a detail.

jamaalscarlett:
@blighj looks like there are some newly failing tests

README.rst
```diff
@@ -460,6 +193,37 @@ JSON example:
...
+
+	device = WebPushDevice.objects.get(registration_id=wp_reg_id)
+
+	title = "Message Received"
+	message = "You've got mail"
+	data = json.dumps({"title": title, "message": message})
+
+	device.send_message(data)
+
+
```
blighj, line 226:
```suggestion
```
This must have been dulplicated during rebasing, and I missed it.

tests/test_rest_framework.py
blighj, line 1: > @blighj looks like there are some newly failing tests

The tests are passing now, this file was failing, it is really outside the scope of this PR. My diagnosis is that a string is more bytes than in py3 than the tests were expecting.

blighj: I've done more digging and I see clearly now why the test was failing, the serilaizer is tied to the model, the registration_id in the APNSDevice has a max_length of 200, and the test was creating a string of lenght 400 ("aE"*200).

So the update I did is appropriate.
@jamaalscarlett This MR is good to go now.

jamaalscarlett: I'm going to re-review and hopefully get this merged today.  Great work.

jamaalscarlett (approved):

blighj:
:) Thank you kindly
I think you could now safely close PR #487 
And the issues #457, #526, #629, #647, #657, and #665

a release to version 3.1.0 would be much appreciated too

PR #677, opened 2022-12-13
fix hardcoded AUTH_USER_MODEL in migrations

joeyorlando:
Hi there 👋 

Great project! I wanted to contribute to fixing a "bug" I realized in the migrations. When you set `USER_MODEL`  as such in your `settings.py`:
```python
PUSH_NOTIFICATIONS_SETTINGS = {
    "USER_MODEL": "user_management.User",
   ...
}
```
the migrations still create a foreign key reference to the value of `AUTH_USER_MODEL`.

This PR would update that to use the value of `settings.PUSH_NOTIFICATION_SETTINGS["USER_MODEL"]`


joeyorlando:
It seems like this may be a duplicate of #572 .. but any way one of these can be merged in and released? 😄

joeyorlando:
closing in favour of #572

PR #689, opened 2023-08-15
Update test.yml

jamaalscarlett:
Fixes failing python 3.6 tests


azmeuk:
@jamaalscarlett python3.6 is unsupported since december 2021
https://endoflife.date/python
Is it necessary to keep it tested?

PR #702, opened 2024-01-16
Add FCM v1 API

ceoy:
Use new FCM v1 API by using firebase_admin SDK.

- Removed GCM type
- Added firebase-admin
- Updated documentation
- Updated tests

Resolves #546

PR #706, opened 2024-02-23
Fix imports of FCM to keep it as an optional dependency

sevdog:
As pointed out in this comment https://github.com/jazzband/django-push-notifications/pull/702#discussion_r1489160018 importing `firebase_admin` or `gcm` at top level in `models.py` turns the `firebase_admin` package as a mandatory requirements while it should not.

To address this I have moved the imports inside the appropriated methods. This allow the codebase to run even without the firebase package (which is optional).

### DIFF
--- a/push_notifications/models.py
+++ b/push_notifications/models.py
@@ -58,9 +58,7 @@ class GCMDeviceManager(models.Manager):
 class GCMDeviceQuerySet(models.query.QuerySet):
 	def send_message(self, message, **kwargs):
 		if self.exists():
-			from firebase_admin import messaging
-
-			from .gcm import dict_to_fcm_message
+			from .gcm import dict_to_fcm_message, messaging
 			from .gcm import send_message as fcm_send_message
 
 			if not isinstance(message, messaging.Message):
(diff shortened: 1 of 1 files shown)


push_notifications/models.py
```diff
@@ -1,9 +1,7 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
-from firebase_admin import messaging
```
50-Course, line 3: This has instead been fixed with #707

50-Course:
Thank you for making this MR. The patch was instead resolving two issues, making `firebase_admin` optional as well as moving the required `.gcm` methods to appropriate area. 

#707 resolved the `firebase_admin` issue, not required. That said, once this conflict herein is fixed. It is good as merged

sevdog:
I have rebased over #707, now this PR makes a single import statement instead of two to handle FCM in models.py.

50-Course:
> I have rebased over #707, now this PR makes a single import statement instead of two to handle FCM in models.py.

I would make one more quick pass, and merge into the `main` trunk. 

Once again, thank you for pulling in this patch.

50-Course (approved):
LGTM 👍🏼

Issue #703, opened 2024-02-16
Problem with setting FIREBASE_APP after migrating to FCM v1


KaiValar:
Hello folks!
I followed the docs about migrating from legacy GCM/FCM to FCM v1 and, when i do the send test notification, it raises a error 500 and this in the log:

`django.core.exceptions.ImproperlyConfigured: Platform FCM, app App-Android does not support the setting: FIREBASE_APP.`

Looks like the FIREBASE_APP setting is not supported, what's wrong here?

Thank you in advance!

### Enviroment
Python 3.11
Django 4.1.11
Django-Push-Notifications: 3.0.2
Firebase-admin 6.4.0

PS: There's a syntax error in de FCM docs, in the part of 
```

# Import the firebase service
from firebase_admin import auth

# Initialize the default app
default_app = firebase_admin.initialize_app()
```

you should import firebase admin for using it.

```

# Import the firebase service
import firebase_admin

# Initialize the default app
default_app = firebase_admin.initialize_app()
```

ceoy:
3.0.2 is still the old version without "new" FCM v1 API.

Good catch with the doc, thank you :) i'll fix it on the weekend (together with https://github.com/jazzband/django-push-notifications/pull/702#discussion_r1489160018).

KaiValar:
Oh, I have migrated ahead of time xD
Thank you, waiting for new version then :)

kmasuhr:
When is the next release scheduled?

PR #707, opened 2024-02-25
make firebase-admin an optional dependency

ceoy:
documentation fix https://github.com/jazzband/django-push-notifications/issues/703
mandatory dependency fix https://github.com/jazzband/django-push-notifications/pull/702#discussion_r1489160018

### DIFF
--- a/push_notifications/models.py
+++ b/push_notifications/models.py
@@ -1,9 +1,7 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
-from firebase_admin import messaging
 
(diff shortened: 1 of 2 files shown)


push_notifications/models.py
```diff
@@ -1,9 +1,7 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
-from firebase_admin import messaging
 
 from .fields import HexIntegerField
-from .gcm import dict_to_fcm_message
```
50-Course, line 6: 👍🏼

PR #699, opened 2024-01-08
Allow default value for cloud message type to be overridden with SETTINGS

kupsum:
Allow default value of ``GCMDevice.cloud_message_type`` to be overridden/specified in ``SETTINGS``.

PR #714, opened 2024-04-10
Add missing migration for the new default FCM choices

jkoestinger:
Should unstuck #699 as well

I noticed that currently if we use the main branch, there is a migration file missing, making Django complain about changes that are not reflected. Here are the changes that are missing.


tuky:
As 699 is (probably rightfully) closed, this is the pr to do now I suggest 😅

jkoestinger:
Sure, I can reopen it :+1:

tuky (approved):

PR #715, opened 2024-04-23
Changed firebase-admin's deprecated send_all method for send_each

### LINKED ISSUE #712: FCM send_all deprecated in favor of send_each/send_each_for_multicast, will stop working in June
Looking through the firebase_admin docs for the [messaging.send_all](https://firebase.google.com/docs/reference/admin/python/firebase_admin.messaging#send_all) method, I saw this: `send_all() is deprecated. Use send_each() instead.`

I investigated a bit further and in the firebase FAQ [found this](https://firebase.google.com/support/faq#fcm-depr-features):

Screenshot 2024-04-08 at 12 21 59

From what I can see in the master branch of the code, all paths from the GCMDevice `send_message` function lead to the `send_all` function from firebase_admin, which would mean that all FCM notifications sent using this library would stop working in June. I know that there has been a big effort for migrating this library to the new FCM V1 API and there is a pending release that's supposed to go out soon but, unless I'm missing something, I think it's worth delaying the release for a few days while this is fixed since it would break FCM functionality if it isn't released before June 21. 

From the docs I gather that `send_each` should basically be a drop in replacement for `send_all`, so the changes wouldn't be big. I can try to make a pull request for these changes, though I'm not sure how to set up and run the tests for the project. 

Let me know if I missed something.

guspix:
Closes #712. Let me know if I need to do anything else.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,7 @@ Dependencies
 - For WebPush (WP), pywebpush 1.3.0+ is required (optional). py-vapid 1.3.0+ is required for generating the WebPush private key; however this
   step does not need to occur on the application server.
 - For Apple Push (APNS), apns2 0.3+ is required (optional).
-- For FCM, firebase-admin 5+ is required (optional).
+- For FCM, firebase-admin 6.2+ is required (optional).
 
 Setup
 -----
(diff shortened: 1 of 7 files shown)


jamaalscarlett:
@guspix I have a PR dropping python 3.6 support https://github.com/jazzband/django-push-notifications/pull/718

jamaalscarlett:
@guspix can I get a review on https://github.com/jazzband/django-push-notifications/pull/718

PR #718, opened 2024-04-26
Remove python 3.6 support.

jamaalscarlett:


guspix:
Looks good to me! Only tiny thing I can see is that the test_admin file has an if statement checking if the Python version is >= 3.7 which I guess would now be innecesary, but that's it. This is my first time contributing so I'm not sure how this works, should I submit a review approving the code changes?

jamaalscarlett:
> Looks good to me! Only tiny thing I can see is that the test_admin file has an if statement checking if the Python version is >= 3.7 which I guess would now be innecesary, but I don't think it really matters. This is my first time contributing so I'm not sure how this works, should I submit a review approving the code changes?

Yes please, once you approve it.  I will merge it.  Then you can rebase your pr to fix the failing 3.6 test.  Then I will cut a release.  Thanks again!

50-Course (approved):
Hello @jamaalscarlet,

thank you for making this patch. Looking at @guspix comment and the updated change int the test,

```python

# 3.6 adds a `,` to the string representation of the exception
python_version = sys.version_info[:2]
if python_version >= (3, 7):
    error_message = "Some messages could not be processed: UnregisteredError('error')"

```

With that, I must say this looks good to me.

jamaalscarlett:
@guspix @50-Course

guspix (approved):

PR #711, opened 2024-03-28
support interruption level on ios devices

50-Course:
Hello team,

Digging deep into `PyAPNs2`, I found that the library does have internal
priority levels, distinctively, `Immediate` and `Delayed`.

In addition to this, support different notification types, such as
`Background`, `Alert`, `VoIP`, `Complication`, `FileProvider` and
`mdm`.

This patch extend one or two of these interfaces to support
`interruption_levels` on User Notification Center. Below is what is
contained in this commit:

- [x] New enumeration types to support interruption levels:
      `time_sensitive`, `active`, `passive`, `critical` alerts
- [x] Wrapper functions to help add additional functionality to existing
      `_apns_send` and `_apns_prepare` methods. This design allows for
      modification of functionality without breaking backwards
      compatibility, and also giving control over what-is-what.
- [x] Functional tests to verify behaviour of this functionality.
      NOTE: This should be tested on real Apple devices or emulators
      before merged.

This patch is a work-in-progress and I would like to get feedback on
this before I proceed with the next steps. Can you take a look at this? @pomali

CC: @jleclanche

Commits:

- **feat: Create module to support library-specific enumeration types**
- **feat(APNS): extend library functionality to support interruption level for Apple Devices**
- **add `interruption_level` to `apns_send_message()` function signature**
- **test(APNs): add functional tests for updated `apns_send_message()` interface**

PR #696, opened 2023-11-02
Add aioapns version of APNS

pomali:
Since version python 3.10+ is not supported by `apns2` (because of dependency on `hyper`) we add `aioapns` version of sending APNs notifications.

We add installation extra `[APNS_ASYNC]` that installs aioapns and use new version of service if `aioapns` is installed. Tests are also conditional on installed modules/version of python.

Related to 
- https://github.com/jazzband/django-push-notifications/pull/623 (is blocked by apns2)
- https://github.com/jazzband/django-push-notifications/issues/685
- https://github.com/jazzband/django-push-notifications/issues/622
- https://github.com/jazzband/django-push-notifications/issues/675

PR #721, opened 2024-05-06
Migrate from apns2 to aioapns

macro1:
Disclaimer: I haven't tested this with real interactions with APNs.

The goal is to remove the indirect use of hyper (which is no longer maintained) and unblock compatibility with newer Python versions without breaking too many of the existing contracts.

I did remove the 'creds' parameter that existed in many of the functions. The value that would have been accepted is a part of apns2, so it cannot be realistically supported.

Somewhat surprising.. It seems the indentation is currently set to tabs? This PR currently converts several of the files, which may not be the right approach.

### DIFF
--- a/push_notifications/apns.py
+++ b/push_notifications/apns.py
@@ -4,141 +4,207 @@ Documentation is available on the iOS Developer Library:
 https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html
 """
 
+import asyncio
 import contextlib
+import tempfile
 import time
 
-from apns2 import client as apns2_client
-from apns2 import credentials as apns2_credentials
-from apns2 import errors as apns2_errors
-from apns2 import payload as apns2_payload
+import aioapns
+from aioapns.common import APNS_RESPONSE_CODE, PRIORITY_HIGH, PRIORITY_NORMAL
+from asgiref.sync import async_to_sync
 
 from . import models
 from .conf import get_manager
-from .exceptions import APNSError, APNSUnsupportedPriority, APNSServerError
+from .exceptions import APNSError, APNSServerError, APNSUnsupportedPriority
+
+
+SUCCESS_RESULT = "Success"
+UNREGISTERED_RESULT = "Unregistered"
 
 
 @contextlib.contextmanager
-def _apns_create_socket(application_id=None):
-	if not get_manager().has_auth_token_creds(application_id):
-		cert = get_manager().get_apns_certificate(application_id)
-		creds = apns2_credentials.CertificateCredentials(cert)
-	else:
-		keyPath, keyId, teamId = get_manager().get_apns_auth_creds(application_id)
-		# No use getting a lifetime because this credential is
-		# ephemeral, but if you're looking at this to see how to
-		# create a credential, you could also pass the lifetime and
-		# algorithm. Neither of those settings are exposed in the
-		# settings API at the moment.
-		creds = apns2_credentials.TokenCredentials(keyPath, keyId, teamId)
-	client = apns2_client.APNsClient(
-		creds,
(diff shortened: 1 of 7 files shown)


macro1:
I closed this when I saw #696, but I'm re-opening it because the approaches are very different. This PR does a full replacement rather than adding an alternate APNS implementation.

50-Course:
i'd be following this development closely, and might have to re-base (#711) on this. Let me know if you need some help or require unblocking at some point.

push_notifications/apns.py
```diff
@@ -4,100 +4,165 @@
...
+	"""
+	client = _apns_create_client(application_id=application_id)
+
+	# if expiration isn't specified use 1 month from now
+	# converting to ttl for underlying library
+	if expiration:
+		time_to_live = expiration - int(time.time())
+	else:
+		time_to_live = 2592000
+
```
50-Course, line 129: nice catch here

push_notifications/apns.py
50-Course, line 1: 👍🏼

push_notifications/apns.py
```diff
@@ -4,100 +4,165 @@
...
+		# settings API at the moment.
+
+	with _apns_path_for_cert(cert) as cert_path:
+		client = aioapns.APNs(
+			client_cert=cert_path,
+			key=key_path,
+			key_id=key_id,
+			team_id=team_id,
+			use_sandbox=get_manager().get_apns_use_sandbox(application_id),
+		)
```
aalbinati, line 58: I found two errors. The topic isn't being sent when creating the client and the context processor raises an exception if there's no credential path. I tested it afterwards and works great with Django 4.2 and Python 3.11

`	if not get_manager().has_auth_token_creds(application_id):
		cert = get_manager().get_apns_certificate(application_id)
		with _apns_path_for_cert(cert) as cert_path:
			client = aioapns.APNs(
				client_cert=cert_path,
				team_id=team_id,
				topic=get_manager().get_apns_topic(application_id),
				use_sandbox=get_manager().get_apns_use_sandbox(application_id),
			)
	else:
		key_path, key_id, team_id = get_manager().get_apns_auth_creds(application_id)
		client = aioapns.APNs(
			key=key_path,
			key_id=key_id,
			team_id=team_id,
			topic=get_manager().get_apns_topic(application_id),
			use_sandbox=get_manager().get_apns_use_sandbox(application_id),
		)`

aalbinati (changes requested):

push_notifications/apns.py
aalbinati: Here's a fork with the fix
[[https://github.com/aalbinati/django-push-notifications/tree/remove-apns2](url)](https://github.com/aalbinati/django-push-notifications/tree/remove-apns2)

macro1: Awesome. Thank you @aalbinati. Doing some testing with the service was on my to-do, I'll pull in the changes you're suggesting soon.

tests/test_apns_models.py
```diff
@@ -13,62 +13,80 @@ class APNSModelTestCase(TestCase):
 
 	def _create_devices(self, devices):
 		for device in devices:
+			print("created", device)
```
HashimJVZ, line 16: You left a print here

pomali:
> I closed this when I saw #696, but I'm re-opening it because the approaches are very different. This PR does a full replacement rather than adding an alternate APNS implementation.

If I remember correctly there were some features not supported by aioapns, and i wanted to avoid breaking changes.

But it looks like async and sync versions have a lot in common so maybe we should refactor async too to share code.

amitc005:
apnsaio released a breaking change where now the key_file for JWT token-based auth is supposed the file content instead of the file path, please ensure you are including that change as well. https://github.com/Fatal1ty/aioapns/pull/62

PR #725, opened 2024-06-03
Fix default FCM_MAX_RECIPIENTS

djch:
Deployed the new release with FCM v1 API support and immediately discovered the default value is too high, triggering a `messages must not contain more than 500 elements` error from the Firebase SDK.

FCM v1 API seems to have changed this limit, so django-push-notifications should adhere to that.


djch:
Doc reference: https://firebase.google.com/docs/cloud-messaging/send-message#send-messages-to-multiple-devices

jamaalscarlett (approved):

PR #734, opened 2024-08-13
modernize distribution builds

dimbleby:
recent releases have been missing wheels eg https://pypi.org/project/django-push-notifications/3.1.0/#files

not sure why: what you are doing in the release workflow is deprecated but so far as I know should work

anyway here is a non-deprecated version


jamaalscarlett:
Thanks for this @dimbleby unfortunately I am not actively using the plugin these days.

PR #738, opened 2024-09-06
Fix error handling on duplicated APNS devices

bielfrontera:
If we have a duplicated registration ID on the APNSDevice model and we get an unregistered error after sending a message, it raises a `MultipleObjectsReturned` exception (get() returned more than one APNSDevice -- it returned ...!).

In order to fix this, we can update all the records with that registration ID instead of using `objects.get`.

### DIFF
--- a/push_notifications/apns.py
+++ b/push_notifications/apns.py
@@ -116,9 +116,7 @@ def apns_send_message(registration_id, alert, application_id=None, creds=None, *
 		)
 	except apns2_errors.APNsException as apns2_exception:
 		if isinstance(apns2_exception, apns2_errors.Unregistered):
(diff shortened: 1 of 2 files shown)


manelclos (approved):

nossila (approved):

jamaalscarlett:
Thanks @bielfrontera. It looks like there is a merge conflict, once that is fixed we can get this merged.

bielfrontera:
> Thanks @bielfrontera. It looks like there is a merge conflict, once that is fixed we can get this merged.

Great! I've rebased and fixed the merge conflict.

50-Course (approved):

Issue #320, opened 2016-09-15
add fields in "aps" (for iOS 10)


Closius:
Talking about [_apns_send()](https://github.com/jleclanche/django-push-notifications/blob/master/push_notifications/apns.py#L143)

To be honest I don't understand why there is no possibilities for manually add custom fields in "aps" sub-dictionary. For example, in iOS 10 a new field "mutable-content" was added in "aps" and there are no ways to add it using django-push-notifications without monkey patching or forking or using "extra" with "aps" key. 

Jleclanche, could you update it?

jleclanche:
I want to rework that function, it shouldn't take a bazillion arguments, should just look in kwargs and have like an aps_extra dict or some such.

HashimJVZ:
Could you let me know if this issue has been resolved?
![image](https://github.com/user-attachments/assets/5aedcde0-2b21-400b-b295-597f0e9d3f45)
Or Do I need to do as the above image?

PR #745, opened 2024-11-06
Add Suppport for mutable_content in aioapns

HashimJVZ:
This PR introduces support for the mutable_content property in the aioapns library, which was previously available only in apns2. The mutable_content property allows notifications to be modified by a Notification Service Extension before being delivered to the user.

This is related to #320 and #350.
Earlier support was available in the `apns.py` file, now added to `apns_async.py`.

`mutable_content` parameter is added to the following methods:
 - apns_send_message
 - apns_send_bulk_message
 Both use a common method `_create_notification_request_from_args` which will take `aps_kwargs`.
 
It will be then added to the NotificationRequest as
```
"aps": {
	"alert": alert,
	"badge": badge,
	"sound": sound,
	"thread-id": thread_id,
	**aps_kwargs,
},
```
 
 Please correct me if I made any misunderstanding. Thanks.

### DIFF
--- a/push_notifications/apns_async.py
+++ b/push_notifications/apns_async.py
@@ -253,6 +253,7 @@ def apns_send_message(
 	loc_key: str = None,
 	priority: int = None,
 	collapse_id: str = None,
+	mutable_content: int = None,
 	err_func: ErrFunc = None,
 ):
 	"""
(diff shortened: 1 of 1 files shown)


50-Course (changes requested):
hello @HashimJVZ thank you for this change. following the recent refactor from #746 can you rebase onto your branch, so this can get reviewed?

PR #746, opened 2024-11-21
Fix #744: Bulk messages with aoiapns not working properly

### LINKED ISSUE #744: Issues with bulk messages in the new aioapns dependency (apns_async)
Hi, i have been trying out the latest async version at scale (Im sending to 200k+ devices in production) and it does not really work for bulk messaging. Already around sending messages to 100 devices it justs hangs indefinitely and never terminates.

I have been trying to fix the issue, however i have zero experience with the underlying libraries (aioapns, h2, aiohttp) and dont really know how its supposed to work. But i managed to make it somewhat stable by using semaphores limiting the number of paralell requests and raising the connection limit. Implementation can be viewed here: https://github.com/mikaelengstrom/django-push-notifications/commit/9a95da876cb25f64d751450a976061b4863c929e

Unfortunatly, performance take quite a hit when done this way.

Very thankful if someone who knows this stuff better could check it out. The original contributor @pomali maybe? :)

mikaelengstrom:
Previous version implemented aioapns in a way that made it hang indefinetly. Especially when receiver list contaned a lot of bad tokens.

Having a lot of bad tokens still affects the reliability and transfer speed of notification sets, which is why this fix also deactivate devices for error codes BadDeviceToken and DeviceTokenNotForTopic unlike previous versions.

Tagging @pomali who initially did the aoiapns integration in case you have any feedback.

### DIFF
--- a/push_notifications/apns_async.py
+++ b/push_notifications/apns_async.py
@@ -1,5 +1,6 @@
 import asyncio
 import time
+
 from dataclasses import asdict, dataclass
 from typing import Awaitable, Callable, Dict, Optional, Union
 
@@ -8,7 +9,7 @@ from aioapns.common import NotificationResult
 
 from . import models
 from .conf import get_manager
-from .exceptions import APNSServerError
+from .exceptions import APNSServerError, APNSError
 
 ErrFunc = Optional[Callable[[NotificationRequest, NotificationResult], Awaitable[None]]]
 """function to proces errors from aioapns send_message"""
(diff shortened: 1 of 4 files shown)


push_notifications/apns_async.py
```diff
@@ -155,105 +126,103 @@ def _create_notification_request_from_args(
...
+		topic: str = None,
+		badge: int = None,
+		sound: str = None,
+		extra: dict = {},
+		expiration: int = None,
+		thread_id: str = None,
+		loc_key: str = None,
+		priority: int = None,
+		collapse_id: str = None,
+		err_func: ErrFunc = None,
```
HashimJVZ, line 225: was this spacing intentional?

mikaelengstrom: Not really, updated with 1 tab spacing for all function arguments

mikaelengstrom:
Would be thankful if for example @50-Course could look into this since you managed the last PR by @pomali

50-Course (commented):
hi @mikaelengstrom. apologies, couldn't make a comment back yesterday - as the notification came in amid few things. i understand the intent behind this change, however, can we have a test suite to ensure this doesn't break things down the line?

mikaelengstrom:
@50-Course I have now fixed the existing testsuite (which were failing when i started). In my opinion it is quite excessive already. Overall i have not changed interface that much, the difference should be that some more statuses is setting "inactive" than before. I also implemented APNSException on failures now to keep the async implementation closer to the original apns implementation. 

Regarding the actual issue i tried to resolve (faulty usage of async causing never exiting state on errors) i do not really see how to implement regression tests for that.

50-Course (approved):

50-Course:
Alright let's call this in!

PR #747, opened 2024-12-20
Add db_index=True for registration_id on APNSDevice model

mikaelengstrom:
When having huge list apns devices without the unique setting activated (UNIQUE_REG_ID) updating devices via the drf-endpoint becomes very slow. Adding an index resolves that.

For reference our current prod-environment have select-statements to this fields taking up to 9 seconds under heavy load.

### DIFF
--- a/push_notifications/models.py
+++ b/push_notifications/models.py
@@ -164,7 +164,9 @@ class APNSDevice(Device):
 		help_text=_("UUID / UIDevice.identifierForVendor()")
 	)
 	registration_id = models.CharField(
-		verbose_name=_("Registration ID"), max_length=200, unique=SETTINGS["UNIQUE_REG_ID"]
+		verbose_name=_("Registration ID"), max_length=200,
+		db_index=not SETTINGS["UNIQUE_REG_ID"],
+		unique=SETTINGS["UNIQUE_REG_ID"],
 	)
 
 	objects = APNSDeviceManager()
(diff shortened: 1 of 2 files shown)


push_notifications/models.py
```diff
@@ -164,7 +164,9 @@ class APNSDevice(Device):
 		help_text=_("UUID / UIDevice.identifierForVendor()")
 	)
 	registration_id = models.CharField(
-		verbose_name=_("Registration ID"), max_length=200, unique=SETTINGS["UNIQUE_REG_ID"]
+		verbose_name=_("Registration ID"), max_length=200,
+		db_index=not SETTINGS["UNIQUE_REG_ID"],
+		unique=SETTINGS["UNIQUE_REG_ID"],
```
50-Course, line 169: Nice catch @mikaelengstrom! This looks great, one more thing, i would like to a profiling request (to have closer look) for this after the holidays before merging. 

For now, approving this PR. thanks.

50-Course (approved):

push_notifications/models.py
50-Course, line 169: one more thing @mikaelengstrom before we call this merged, 

here we seem to be resolving `unique=False` given the constant, `UNIQUE_REG_ID` defaults to false. is this intentional? 

happy new year

50-Course (changes requested):

push_notifications/models.py
mikaelengstrom: @50-Course Yes. So before this change, the unique attribute is set based on the UNIQUE_REG_ID boolean setting, which is implicitly creating an index. However, for those of running the default value for `UNIQUE_REG_ID` (False), no index is created making it very inperformant.

So in essence, after this change, either the `db_index` or the `unique` attribute will be set depending on the users settings. Ensuring that some kind of index is created.

50-Course: right got it, thanks!

50-Course (approved):

PR #755, opened 2025-01-11
Add Support for mutable_content in aioapns

HashimJVZ:
Re-do of #745 

Changes in the #745 was https://github.com/jazzband/django-push-notifications/commit/314a48410362f18f766608753a93248f7fc0eba6

This PR introduces support for the mutable_content property in the aioapns library, which was previously available only in apns2. The mutable_content property allows notifications to be modified by a Notification Service Extension before being delivered to the user.

This is related to https://github.com/jazzband/django-push-notifications/issues/320 and https://github.com/jazzband/django-push-notifications/pull/350.
Earlier support was available in the apns.py file, now added to apns_async.py.

### DIFF
--- a/push_notifications/apns_async.py
+++ b/push_notifications/apns_async.py
@@ -222,6 +222,7 @@ def apns_send_message(
 	loc_key: str = None,
 	priority: int = None,
 	collapse_id: str = None,
+	mutable_content: int = None,
 	err_func: ErrFunc = None,
 ):
 	"""
@@ -253,6 +254,7 @@ def apns_send_message(
 		loc_key=loc_key,
 		priority=priority,
 		collapse_id=collapse_id,
+		mutable_content=mutable_content,
 		err_func=err_func,
 	)
 
@@ -277,6 +279,7 @@ def apns_send_bulk_message(
 	loc_key: str = None,
 	priority: int = None,
 	collapse_id: str = None,
+	mutable_content: int = None,
 	err_func: ErrFunc = None,
 ):
 	"""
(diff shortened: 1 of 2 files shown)


50-Course (commented):
Hi @HashimJVZ,

This is great, I had some review in the past four days -- however, i couldn't fully complete it due to one or two. Going through this again, I have one or two questions -- so we can call this in 😄 . 

Foremost, this looks great! 

https://github.com/jazzband/django-push-notifications/blob/e0e098cbeba8dbee0502c046f2ad6e64d0ed87fe/tests/test_apns_async_push_payload.py#L185-L208

however, we are introducing `mutable_content` as `int` (rightfully so) in this change, as opposed to previous versions built upon `apns2` library (`mutable_content` as `boolean` - which does the handling underneath, constraining the top-level API to boolean values), i think the former anti-pattern with the boolean values is better here, as it avoid the unnecessary shenanigans with magic numbers from user inputs. However, we with good doc, we good. Can we document this new change? 

It would help smooth the smooth transition in the release note. Other than that, this looks ready for join the mainline 👍🏼

HashimJVZ:
The API still accepts mutable_content=True as a boolean, and we only convert it to 1 when building the payload for APNs, following Apple’s convention. The caller doesn’t need to pass integers explicitly.

50-Course:
Hey @HashimJVZ, yes the docstring would suffice. somehow I got the notifications of your reply, but can't find it out here.

HashimJVZ:
@50-Course , I used int instead of bool as type hinting by mistateke. I just realised my mistake in the code after commenting. So I deleted immediately.

50-Course (approved):
this is now marked resolved, let's get this out there 🚀

PR #758, opened 2025-01-18
Add category to apns-async calls

aaronn:
APNS has the concept of a `category` and this was lost in the addition of the async version of this package with the addition of `apns-async`. This re-adds category support.


aaronn:
This doesn't work

50-Course (commented):
oh i see, this should have been part of `aps_kwargs`, @HashimJVZ change proposed a potential fix for this (#760).

PR #763, opened 2025-02-18
Make mutable-content int (1) instead of bool (True)

HashimJVZ:
APNS expect mutable-content as int in the message payload.
Setting `mutable_content` to True, will set `mutable-content` to 1.

### DIFF
--- a/push_notifications/apns_async.py
+++ b/push_notifications/apns_async.py
@@ -239,7 +239,7 @@ def apns_send_message(
 	:param alert: The alert message to send
 	:param application_id: The application_id to use
 	:param creds: The credentials to use
-	:param mutable_content: If True, enables the "mutable-content" flag in the payload.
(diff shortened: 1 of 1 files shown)


push_notifications/apns_async.py
```diff
@@ -397,10 +397,9 @@ async def _send_bulk_request(
 
 	aps_kwargs = {}
 	if mutable_content:
-		aps_kwargs["mutable-content"] = mutable_content
+		aps_kwargs["mutable-content"] = 1
```
50-Course, line 400: right this is to help align with the implementation of apns.py?

50-Course (approved):

push_notifications/apns_async.py
50-Course: Update: this is looks great! Just confirmed with the tests, no additional change required. Thanks for the MR 👍🏼

Issue #662, opened 2022-06-29
running migration problem


ahmedelfateh:
after setting up django_push_notification and adding push_notification to the app, I run the migration and this msg appere 

```
Your models in app(s): 'push_notifications' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```
  
 and when run makemigrations, this  appere
 
```
 Migrations for 'push_notifications':
  /usr/local/lib/python3.8/site-packages/push_notifications/migrations/0010_auto_20220629_1231.py
    - Alter field id on apnsdevice
    - Alter field id on gcmdevice
    - Alter field id on webpushdevice
    - Alter field id on wnsdevice
```
    
I am using the django~=3.2 / django-push-notifications==3.0.0 
with python:3.8.12

what is the problem here?

jamaalscarlett:
@ahmedelfateh This is happening because of https://docs.djangoproject.com/en/3.2/ref/settings/#default-auto-field. If you started your project on Django 3.2 or newer, your settings.py file contains 
```
DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
```
but django push notifications uses AutoField.  The fix would be to change the setting to:
```
DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
```
or except the new migration.

denizdogan:
@jamaalscarlett Is there any reason why this migration can't just be included in the library at this point? Django 2 is not supported any longer, 3.2 is the oldest supported version, and it's LTS until April 2024.

PureTryOut:
I too would prefer if the library just included the migration. I have no interest in a Django version lower than 3.2 and now every time I run `makemigrations` this migration comes up I have to remove (as it's not in my version control and thus not persistent).

jamaalscarlett:
@PureTryOut I will take a look at this tonight

Aidar:
Any updates?

CoderMungan:
any update?

CoderMungan:
@diesieben07 same way i solved problem. Thanks! Can we put this code in the library and pull it directly from the library if @jamaalscarlett allows?

diesieben07:
Sure, yes please!

VctorAHernndez:
[...]

PR #769, opened 2025-05-27
Add explicit Django AppConfig to set default_auto_field

diesieben07:
Django projects might set `DEFAULT_AUTO_FIELD` to different values than `AutoField`.
In fact modern Django versions generate a settings.py file that uses `BigAutoField` by default.
The result is that when using this library in such a project, `makemigrations` will always attempt to create migrations for push_notification's models.

By adding an explicit `AppConfig`, Django knows that this application uses `AutoField`, regardless of the `DEFAULT_AUTO_FIELD` setting.

See:
- https://docs.djangoproject.com/en/5.2/ref/applications/#django.apps.AppConfig.default_auto_field
- https://docs.djangoproject.com/en/5.2/ref/applications/#configuring-applications

Fixes https://github.com/jazzband/django-push-notifications/issues/662

### DIFF
--- a/push_notifications/apps.py
+++ b/push_notifications/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class PushNotificationsConfig(AppConfig):
+	name = "push_notifications"
+	default_auto_field = "django.db.models.AutoField"


50-Course (approved):

50-Course (approved):
Hi @diesieben07, thank you for this patch. Given the importance of this, we should pull this in with priority

PR #772, opened 2025-06-03
Softcoding APNS timeout

goinnn:
Setting a timeout for requests is crucial to prevent an application from hanging indefinitely while waiting for a response. It enhances system resilience by allowing detection of failures or slowdowns in external services. Additionally, having a configurable timeout makes it adaptable to different performance and reliability needs.


50-Course:
@goinnn can you please take a look at the conflicts? this should have gone in long ago, was expecting to see some other approvals before it goes in. Thank you for your contributions to help make this better for everyone

PR #773, opened 2025-06-03
Fix Edge pushes: TTL configurable 

goinnn:
This implementation makes the TTL configurable A TTL is required for Edge; without it, a 400 Bad Request error is returned.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,7 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
 - ``WP_PRIVATE_KEY``: Absolute path to your private certificate file: os.path.join(BASE_DIR, "private_key.pem")
 - ``WP_CLAIMS``: Dictionary with default value for the sub, (subject), sent to the webpush service, This would be used by the service if they needed to reach out to you (the sender). Could be a url or mailto e.g. {'sub': "mailto:development@example.com"}.
 - ``WP_ERROR_TIMEOUT``: The timeout on WebPush POSTs. (Optional)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -130,6 +130,7 @@ For WNS, you need both the ``WNS_PACKAGE_SECURITY_KEY`` and the ``WNS_SECRET_KEY
 - ``WP_PRIVATE_KEY``: Absolute path to your private certificate file: os.path.join(BASE_DIR, "private_key.pem")
 - ``WP_CLAIMS``: Dictionary with default value for the sub, (subject), sent to the webpush service, This would be used by the service if they needed to reach out to you (the sender). Could be a url or mailto e.g. {'sub': "mailto:development@example.com"}.
 - ``WP_ERROR_TIMEOUT``: The timeout on WebPush POSTs. (Optional)
+- ``WP_TTL``: Time to live: The maximum amount of time (in seconds) a push message should be kept in the messaging service's queue before it expires, if it hasn’t been delivered to the device.. (Optional, default value is 5 minutes, 300 seconds)
```
50-Course, line 133: excellent, may i inquire if this differs from `TTL header`, i am assuming its the same?

> The TTL Header is the number of seconds the notification should stay in storage if the remote user agent isn’t actively connected. “0” (Zed/Zero) means that the notification is discarded immediately if the remote user agent is not connected; this is the default. This header must be specified, even if the value is “0”.


Ref: https://blog.mozilla.org/services/2016/08/23/sending-vapid-identified-webpush-notifications-via-mozillas-push-service/

goinnn: pywebpush will use WP_TTS  to TTL header:

https://github.com/web-push-libs/pywebpush/blob/main/pywebpush/__init__.py#L393

By default ttl is 0. And this is not a value accepted by Edge

50-Course (approved):
This is good to go!

50-Course:
just realize we still don't have second approval on this, please take a look at this too, if you can, let's merge this in 🙏🏼

PR #793, opened 2025-12-14
Add workaround for `py-vapid` cryptography deprecation warning

### LINKED ISSUE #781: vapid fails due to dependency issue (includes solution)
I tried to generate keys, as documented on [docs/WebPush](https://github.com/jazzband/django-push-notifications/blob/master/docs/WebPush.rst) but got stuck because of this issue: https://github.com/web-push-libs/vapid/issues/105

```
/usr/local/lib/python3.13/site-packages/py_vapid/__init__.py:200: CryptographyDeprecationWarning: Curve argument must be an instance of an EllipticCurve class. Did you pass a class by mistake? This will be an exception in a future version of cryptography
```

**The solution** a simpler way to generate vapid keys use this script is here: https://gist.github.com/Tobiaqs/450a4516ae44813792b7d84028c366c0

Maybe this could be solved just with a docs update?

50-Course:
Updates WebPush documentation to provide a working alternative for VAPID key generation while the upstream `py-vapid` issue is being resolved.

#### References

- Upstream issue: https://github.com/web-push-libs/vapid/issues/105
- Alternative script: https://gist.github.com/Tobiaqs/450a4516ae44813792b7d84028c366c0

Fixes #781

### DIFF
--- a/docs/WebPush.rst
+++ b/docs/WebPush.rst
@@ -7,33 +7,76 @@ These are in addition to the instalation steps for django-push-notifications[WP]
 
 Configure the VAPID keys
 ------------------------------
-- Install:
 
-.. code-block:: python
+.. note::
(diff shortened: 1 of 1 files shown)


50-Course:
@jamaalscarlett can we give this a priority approval?

Issue #750, opened 2025-01-08
issues with apns_async


valhallen13:
Hi,

We send notifications to thousands of users, and there are a lot of TimeoutErrors being thrown by the library. I have looked through your code quite a bit and have noticed 2 strange things.
- apns_async.py:383 the timeout is hardcoded to 1 second. I think this should be a property that can be set in the settings.
- apns_async.py:336 i dont see the point of raising an error, because the errors are already available in the "results" variable. Raising an error just interrupts the whole flow, and does not even return the results variable which would allow us to mark users that have already received the message. 
 
Furthermore, sending to thousands of users means that there will always be at least one registration_id that will be Unregistered, so the code will throw an error every single time. I dont think that having to run this code in a perpetual try:except should be the default way to implement it. It would work much smoother if you do not raise the error on line 336, or if you return the entire list of results with the exception so at least we can mark the successful users.

pomali:
About (1) (hardcoded timeout) - would it make sense for you to have option of overriding `_create_client()` or is that overkill?


About (2) (raising errors) - if I remember correctly raising errors is how `apns.py` (non-async) version works because of thats how `apns2` handles things. I agree with you that it's cumbersome when handling bulk requests. So you think not raising error when we get unsuccessful result is the best way?

igor-wl:
I would like to +1 for the mentioned issues, especially the second one. What we are seeing is every time there is at least one unregistered user we get an exception from `send_message` and no way of knowing which users got the message.

Essentially if everything was successful we get `results` dict with "Success" state for all and in any other case we don't have access to the results dict but instead a single error message with concatenated errors which we can't even match to the registration/user. What is the purpose of `results` if we can access it only when EVERYTHING is successful?

The way I see it we need to:
1. Remove `raise APNSError` because it breaks the flow and completely disrupts usefulness of returning separate result for each user.
2. Update conditions because sync version of bulk message checks only for `Unregistered` message, while async version checks for `Unregistered`, `BadDeviceToken`, and `DeviceTokenNotForTopic`.

I read the comments on the original issue #744 and what @pomali said about `apns2` handling things that way, but I just don't get the reasoning for raising an error and what exactly `apns2` does that way?

s4cha:
+1 on @valhallen13 remarks, exceptions are "exceptions", in this specific case of having 1 unregistered device or one Timeout in a big batch is fairly common. I'd err on the side of returning a "report" object without raising an exception.

[...]

PR #794, opened 2025-12-14
fix: make APNS async timeout configurable

50-Course:
This is the first of two PRs addressing #750. This PR specifically tackles the hardcoded timeout issue in `apns_async.py`.

In this PR, I've replaced the hardcoded timeout values with configurable parameters. Users can now set their desired timeout values when initializing the APNs client. This change enhances flexibility and allows for better handling of different network conditions.

### Changes Made

- Introduced a `timeout` parameter in the APNs client initialization.
- Updated all instances where the hardcoded timeout was used to reference the new parameter.
- Added default timeout values to maintain backward compatibility.

#### Notes

This PR intentionally does NOT address the error handling issue - that will come in a follow-up PR to keep changes focused and reviewable.

cc @valhallen13 @igor-wl @s4cha


igor-wl:
You'll probably want to mention the new config `APNS_ERROR_TIMEOUT` in the [README.rst](https://github.com/jazzband/django-push-notifications?tab=readme-ov-file#settings-list) file

PR #798, opened 2026-03-25
Fix: omit alert key from APS payload when alert=None (silent notification)

yuekui:
When `alert=None` was passed to `apns_send_message()` or `apns_send_bulk_message()`,
the code silently converted it to `Alert(body="")` and still included `"alert": {"body": ""}`
in the APS payload. This made true silent notifications impossible via this API.

The docstring documented `alert=None` as the way to send silent notifications, but the
implementation did not match.

### DIFF
--- a/push_notifications/apns_async.py
+++ b/push_notifications/apns_async.py
@@ -116,7 +116,7 @@ class Alert:
 
 def _create_notification_request_from_args(
 	registration_id: str,
-	alert: Union[str, Alert, None],
+	alert: Optional[Union[str, Alert]],
 	badge: Optional[int] = None,
 	sound: Optional[str] = None,
 	extra: Optional[Dict[str, Any]] = None,
@@ -223,7 +223,7 @@ def _get_credentials(application_id: Optional[str] = None) -> Credentials:
 
 def apns_send_message(
 	registration_id: str,
-	alert: Union[str, Alert, None],
+	alert: Optional[Union[str, Alert]],
 	application_id: Optional[str] = None,
 	creds: Optional[Credentials] = None,
 	topic: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


push_notifications/apns_async.py
```diff
@@ -116,7 +116,7 @@ def asDict(self) -> Dict[str, Any]:
 
 def _create_notification_request_from_args(
 	registration_id: str,
-	alert: Union[str, Alert],
+	alert: Union[str, Alert, None],
```
jamaalscarlett:
```suggestion
	alert: Optional[Union[str, Alert]],
```

push_notifications/apns_async.py
```diff
@@ -220,7 +223,7 @@ def _get_credentials(application_id: Optional[str] = None) -> Credentials:
 
 def apns_send_message(
 	registration_id: str,
-	alert: Union[str, Alert],
+	alert: Union[str, Alert, None],
```
jamaalscarlett:
```suggestion
	alert: Optional[Union[str, Alert]],
```

push_notifications/apns_async.py
```diff
@@ -289,7 +292,7 @@ def apns_send_message(
 
 def apns_send_bulk_message(
 	registration_ids: list[str],
-	alert: Union[str, Alert],
+	alert: Union[str, Alert, None],
```
jamaalscarlett:
```suggestion
	alert: Optional[Union[str, Alert]],
```

push_notifications/apns_async.py
```diff
@@ -386,7 +389,7 @@ def apns_send_bulk_message(
 
 async def _send_bulk_request(
 	registration_ids: list[str],
-	alert: Union[str, Alert],
+	alert: Union[str, Alert, None],
```
jamaalscarlett:
```suggestion
	alert: Optional[Union[str, Alert]],
```

jamaalscarlett (changes requested):
Small suggestion, looks good. Thanks!

yuekui:
> Small suggestion, looks good. Thanks!

Applied, thanks

